PR #51, opened 2020-08-28
fix: recursive folder creation

YuryShkoda:


krishna-acondy (commented):
It'd be good if we can pull in the changes from main, and fix the formatting according to the prettier config. Other than that, it looks good! 🙌🏽

allanbowe (approved):

PR #70, opened 2020-09-08
This closes #67 , closes #68 and closes #69

### LINKED ISSUE #67: CLI crashes on run command
I expected this to crash as I hadn't `sasjs add`'d however the message should have been more helpful

[asciicast](https://asciinema.org/a/j4ii3d4OmHoyAbTRCpk0d2GXX)

Definition of done:  helpful error messages when attempting to run code in a misconfigured state

### LINKED ISSUE #68: sasjs create SASONLY template
We need a template for adding just the SAS component to a frontend project.

the command will be `sasjs create [name] -t sasonly`

* if [name] is missing, it should be created in the current directory (this should be true for the other templates also)
* the folowing sasjs subfolders will be created:  [macros,build,db,services]
* the sasjs/services/common/appinit.sas will be created
* npm install @sasjs/core will execute
* the .gitignore will be created or updated
* a minimal sasjsconfig.json will be created, with an empty targets array

### LINKED ISSUE #69: sasjs add (local) does not work if local env does not exist
If the user wishes to add a local config, but he is not in an existing repository, we should deploy a "sasonly" template (https://github.com/sasjs/cli/issues/68) to avoid the situation below

image

saadjutt01:


src/cli.js
```diff
@@ -262,39 +263,54 @@ function processRunParameters(parameters) {
...
-        `Invalid web app type.\nSupported types are ${chalk.cyanBright(
-          'angular'
-        )}, ${chalk.cyanBright('react')} and ${chalk.cyanBright('minimal')}.`
-      )
+function invalidAppType() {
+  console.error(
+    chalk.redBright(
+      `Invalid web app type.\nSupported types are ${chalk.cyanBright(
+        'angular'
+      )}, ${chalk.cyanBright('react')} and ${chalk.cyanBright('minimal')}.`
```
allanbowe, line 284: we need sasonly here as well right?

allanbowe (approved):

src/cli.js
saadjutt01: got it

krishna-acondy (approved):
Looks good to me!

PR #83, opened 2020-09-11
feat: execution of SAS job

### LINKED ISSUE #80: sasjs request feature
We need a new command:

`sasjs request`

This is in order to run existing, deployed SAS services.  It will take the same params as the adapter, service location, data (path to a json file), and config (if overriding).

We will then get as output the following files:
```
[servicename].json -> the output of the service
[servicename].log -> the log
```
 This output will go in the `sasjsbuild` folder if running in a local project, else the current directory if global.  At a later date we should set a default in the config for these results (and same for sasjs run)

medjedovicm:
Closes #80


src/sasjs-request/index.js
```diff
@@ -0,0 +1,103 @@
...
+        chalk.redBright(`Login callback called. Request failed.`)
+      },
+      accessToken
+    )
+    .then(
+      (res) => {
+        console.log('Request finish', res)
+      },
+      (err) => {
+        chalk.redBright(`Login callback called. Request failed. ${err}`)
```
krishna-acondy, line 100: @YuryShkoda has added a `displayResult` function in the context command which contains some extra logic for displaying errors better. Maybe we could pull that out into its own file and use it here?

YuryShkoda: I already moved displayResult() to a separate utility. Will be implemented in the upcoming PR.

YuryShkoda (approved):
looks good to me

allanbowe (dismissed):

src/sasjs-help/index.js
```diff
@@ -150,6 +150,10 @@ export async function printHelpText() {
     )} - Lets the user run a given SAS file against a specified target.
     The target can exist either in the local project configuration or in the global .sasjsrc file.
 
+    * ${chalk.greenBright(
+      'request <sasJob> -d <path/to/datafile> -c <path/to/configfile> -t <targetName>`.'
+    )} - Lets the user run a SAS job against a specified target.
+    The target can exist either in the local project configuration or in the global .sasjsrc file.
```
allanbowe, line 156: `<sasJob>` is confusing - is it a UUID, a job name?  Actually it is the full path and name.

Presumably `-d`, `-c` and `-t` are all optional?

allanbowe (approved):

src/sasjs-request/index.js
```diff
@@ -0,0 +1,140 @@
...
+        try {
+          output = JSON.stringify(res, null, 2)
+        } catch (error) {
+          displayResult(null, null, 'Request finished.')
+
+          return
+        }
+
+        let outputPath = path.join(
+          process.env.PWD,
```
krishna-acondy, line 114: We might want to use `process.cwd()` here since it will work both on Windows and Unix.

https://stackoverflow.com/questions/30991319/whats-the-difference-between-process-cwd-vs-process-env-pwd

src/sasjs-request/index.js
```diff
@@ -0,0 +1,140 @@
...
+        try {
+          output = JSON.stringify(res, null, 2)
+        } catch (error) {
+          displayResult(null, null, 'Request finished.')
+
+          return
+        }
+
+        let outputPath = path.join(
+          process.env.CWD,
```
krishna-acondy, line 114: I believe this needs to be `process.cwd()` instead.

allanbowe (approved):

PR #87, opened 2020-09-15
fix(sasjs-context): read access token from environment

krishna-acondy:
* Read `access_token` from environment if it wasn't found in the target configuration.
* Improve error parsing logic - this currently fails when the returned error body is not JSON.

We now show the parse error along with the original error like so:
image

### DIFF
--- a/src/utils/config-utils.js
+++ b/src/utils/config-utils.js
@@ -216,3 +216,17 @@ export async function getProjectRoot() {
   }
   return root
 }
+
+export function getAccessToken(target) {
+  const accessToken =
+    target && target.authInfo && target.authInfo.access_token
+      ? target.authInfo.access_token
+      : process.env.access_token
+
+  if (!accessToken || accessToken.trim() === 'null') {
+    throw new Error(`A valid access token was not found.
+    Please provide an access token in the access_token property in your .env file or as part of the authInfo in your target configuration (sasjsconfig.json).`)
+  }
+
+  return accessToken
+}
--- a/test/config-utils.spec.js
+++ b/test/config-utils.spec.js
@@ -5,6 +5,10 @@ describe('Config Utils', () => {
     process.env.access_token = null
   })
 
+  afterEach(() => {
+    process.env.access_token = null
+  })
+
   describe('getAccessToken', () => {
     test('should get access token from authInfo', () => {
       const target = {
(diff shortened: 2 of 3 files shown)


YuryShkoda (approved):

allanbowe (dismissed):

test/config-utils.spec.js
```diff
@@ -0,0 +1,50 @@
...
+          access_token: 'T0K3N'
+        }
+      }
+
+      const token = getAccessToken(target)
+
+      expect(token).toEqual('T0K3N')
+    })
+
+    test('should throw an error when access token is unavailable', () => {
```
YuryShkoda, line 25: In this test case, access token is available, but not valid.

YuryShkoda (changes requested):

test/config-utils.spec.js
krishna-acondy: I’d still leave it as unavailable, since I think the empty string means the value hasn’t been specified.
To me, invalid means that the access token can’t be parsed, or that it has expired, which we are not checking for here.
But happy to change it if we feel really strongly about it as well. :slightly_smiling_face:

src/utils/config-utils.js
```diff
@@ -216,3 +216,17 @@ export async function getProjectRoot() {
...
+
+export function getAccessToken(target) {
+  const accessToken =
+    target && target.authInfo && target.authInfo.access_token
+      ? target.authInfo.access_token
+      : process.env.access_token
+
+  if (!accessToken || accessToken.trim() === 'null') {
+    throw new Error(`A valid access token was not found.
+    Please provide an access token in the access_token property in your .env file or as part of the authInfo in your target configuration (sasjsconfig.json).`)
```
allanbowe, line 228: are we still supporting the sasjsconfig approach?

test/config-utils.spec.js
allanbowe: empty / undefined / null I would say all mean 'unavailable' (in the sense we're sure a valid token is not available) happy to be overruled though

allanbowe (approved):

src/utils/config-utils.js
krishna-acondy: We are, yes. Shouldn't we?

YuryShkoda (approved):

src/utils/config-utils.js
allanbowe: didn't see this.  No, we shouldn't

krishna-acondy: Slightly too late. Could you create a separate issue for this please?

PR #89, opened 2020-09-16
chore(context): changed 'context' command to work in batch

YuryShkoda:


medjedovicm (approved):

src/sasjs-help/index.js
```diff
@@ -101,18 +101,15 @@ export async function printHelpText() {
...
+          example: sasjs create myTarget -source ../contextConfig.json
         * ${chalk.cyanBright('edit')} - edits existing context.
-          1) Provide target name.
-          2) Provide path to context config file (eg. ./contextConfig.json).
-        * ${chalk.cyanBright('remove')} - removes existing context.
-          1) Provide target name.
-          2) Provide path to context config file (eg. ./contextConfig.json).
+          example: sasjs edit myTarget -source ../contextConfig.json
+        * ${chalk.cyanBright('delete')} - deletes existing context.
+          example: sasjs delete myTarget -source ../contextConfig.json
```
allanbowe, line 108: why is the `-source` parameter needed here?

also, can we provide an example of the required JSON structure in the help?

src/sasjs-help/index.js
```diff
@@ -101,18 +101,15 @@ export async function printHelpText() {
       'context <command>'
     )} - performs operations on contexts.
         * ${chalk.cyanBright('create')} - creates new context.
-          1) Provide target name.
-          2) Provide path to context config file (eg. ./contextConfig.json).
+          example: sasjs create myTarget -source ../contextConfig.json
```
allanbowe, line 104: myTarget is a misleading context name - in SASjs nomenclature, a target is a SAS server.  

which is a good point, we should add a target parameter, and default to the first target if a target is not found in the sasjsconfig file

allanbowe (approved):

allanbowe:
Didn't mean to hit approve - the documentation needs improving, also the part about the target needs clarifying.

Finally - has this been tested with a global (not just local) config?

src/sasjs-help/index.js
YuryShkoda: fixed

src/sasjs-help/index.js
YuryShkoda: fixed

src/sasjs-help/index.js
```diff
@@ -101,18 +101,38 @@ export async function printHelpText() {
...
+
+          context config example:
+            {
+              "name": "New context",
+              "launchName": "SAS Job Execution launcher context",
+              "sharedAccountId": "cas",
+              "autoExecLines": ["%put hello;"],
+              "authorizedUsers": []
+            }
+          NOTE: Keys "name", "launchName", "sharedAccountId", "autoExecLines", "authorizedUsers" are required.
```
allanbowe, line 115: are we sure that autoExecLines is required?

allanbowe (approved):

src/sasjs-context/index.js
```diff
@@ -16,27 +16,74 @@ export async function processContext(command) {
...
+        ).join('\n')}`
+      )
+    )
 
     return
   }
 
-  const targetName = await getAndValidateTargetName()
+  const commandExample =
+    'sasjs context <command> -source ../contextConfig.json -target targetName'
```
medjedovicm, line 31: When passing parameters that are more then one letter, should it be prepended with 2 dashes? `--`

src/sasjs-help/index.js
YuryShkoda: according addapter's function, yes
image

src/sasjs-context/index.js
YuryShkoda: @allanbowe should I change it to use `--`?

YuryShkoda: fixed

PR #95, opened 2020-09-21
fix(web): added encoding to base64 *.js and *.css files for SAS9 targets

### LINKED ISSUE #93: sasjs stream - base64 encode JS & CSS on SAS 9 only
When `streamWeb:true` and `serverType` is SAS9 we should base64 encode the JS and CSS files into SAS.

This is because JS and CSS files can contain UTF-8 characters that are not supported in customer environments that use WLATIN1.

For Viya, the encoding is always UTF-8 so the conversion is not needed.

The downside of the conversion is much higher file sizes (4x) however these assets are cached by the browser so the impact is only felt once on startup.

We should invoke these using JS64 and CSS64 in the function, eg:

```
%sasjsout(JS64)
```

YuryShkoda:
closes #93


krishna-acondy (approved):

allanbowe (changes requested):
Just one part left - when we _invoke_ `sasout()` we need to pass in JS64 or CSS64 instead of JS / CSS (SAS 9 only)

YuryShkoda:
> Just one part left - when we _invoke_ `sasout()` we need to pass in JS64 or CSS64 instead of JS / CSS (SAS 9 only)

fixed

krishna-acondy (dismissed):

allanbowe (approved):

PR #98, opened 2020-09-21
fix: request use accessToken from .env

medjedovicm:


src/sasjs-request/index.js
```diff
@@ -8,7 +8,7 @@ import {
   createFolder
 } from '../utils/file-utils'
 import { getVariable } from '../utils/utils'
-import { getAccessToken } from '../utils/auth-utils'
+import { getAccessToken } from '../utils/config-utils'
```
krishna-acondy, line 11: The method signatures are different, so this won't work without changing the call below.

krishna-acondy (changes requested):

src/sasjs-request/index.js
```diff
@@ -74,13 +74,7 @@ export async function runSasJob(
...
   }
 
-  const accessToken = await getAccessToken(
-    sasjs,
-    clientId,
-    clientSecret,
-    target,
-    isLocal
-  )
+  const accessToken = await getAccessToken(target)
```
krishna-acondy, line 77: We don't need the bits above this where we're getting the `clientId` and `clientSecret` anymore.

src/sasjs-request/index.js
```diff
@@ -59,28 +58,7 @@ export async function runSasJob(
...
-  }
-
-  const accessToken = await getAccessToken(
-    sasjs,
-    clientId,
-    clientSecret,
-    target,
-    isLocal
-  )
+  const accessToken = await getAccessToken(target)
```
krishna-acondy, line 61: No need for `await` since `getAccessToken` is not async.

krishna-acondy (approved):

PR #212, opened 2020-11-07
fix: folder regression, services stay in the root

### LINKED ISSUE #210: fix folder regression
since v1.18 the folder setup has been modified.

Instead of:

`Appname/folder/servicename`

We now have

`Appname/services/folder/servicename`

This will break all apps that use the adapter with the standard `appLoc/folder/service` format.

Expected behaviour:

**services** are deployed in the old fashion, where each direct subfolder contains the services

**jobs** are deployed in a jobs folder, with corresponding sub(sub) folders.

This will cause a clash if services are created in a folder called `jobs` but that is a fairly unlikely scenario, and we can just reserve that as a folder name.

medjedovicm:
## Issue

Closes #210 

## Intent

Leave the services in root, instead of in subfolder.

## Implementation

Changed the code that is creating json file content.

## Checks

Someone else must run those, my environment doesn't work.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -6856,7 +6856,7 @@
       "dependencies": {
         "JSONStream": {
           "version": "1.3.5",
-          "resolved": "https://registry.npmjs.org/JSONStream/-/JSONStream-1.3.5.tgz",
+          "resolved": false,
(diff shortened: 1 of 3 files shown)


allanbowe (approved):

allanbowe:
failing test ignored as relates to a seperate issue

PR #223, opened 2020-11-11
fix: sasjs job execute issue for git-bash windows

### LINKED ISSUE #221: sasjs job execute issue
for some reason the `sasjs job execute` command is adding a _local_ filepath to the provided path.

SAS jobs _always_ execute on the server, so wouldn't expect this.

Also, the message says it has been submitted but that can't be correct as in fact it wasn't found:

image

saadjutt01:
## Issue

Closes #221 

## Intent

Bash breaks Windows tools by replacing forward slash with a directory path
POSIX path conversion: http://www.mingw.org/wiki/Posix_path_conversion

## Implementation

Written regex to replace if /git/ path is pre appended in args in cli.js file

## Checks

### DIFF
--- a/src/cli.js
+++ b/src/cli.js
@@ -25,7 +25,8 @@ import { exit } from 'process'
 function parseCommand(rawArgs) {
   checkNodeVersion()
 
-  const args = rawArgs.slice(2)
+  // should update paths converted by POSIX in case of process.env.MSYSTEM is present
+  const args = rawArgs.slice(2).map((arg) => arg.replace(/(.*)\/Git\//i, '/'))
 
   if (args.length) {
     const name = getUnaliasedCommand(args[0])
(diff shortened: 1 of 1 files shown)


src/cli.js
```diff
@@ -25,7 +25,8 @@ import { exit } from 'process'
 function parseCommand(rawArgs) {
   checkNodeVersion()
 
-  const args = rawArgs.slice(2)
+  // should update paths converted by POSIX in case of process.env.MSYSTEM is present
+  const args = rawArgs.slice(2).map((arg) => arg.replace(/(.*)\/Git\//i, '/'))
```
allanbowe, line 29: My concerns here are:
1. What if the current folder does not end with "Git"
2. What if the SAS Drive folder contains the word "Git"

Can we somehow extract that renegade local directory, and use it in the regex?

allanbowe (approved):

PR #227, opened 2020-11-12
fix: deploy is forced by default

### LINKED ISSUE #211: `sasjs deploy` should always force deploy
when running `sasjs deploy`, the expectation is always that the `appLoc` will be deleted and rebuilt, so the `-f` flag should be applied by default

image

medjedovicm:
## Issue

Closes #211 

## Intent

Deploy is always forced, by default.

## Checks


src/main.js
```diff
@@ -111,6 +111,11 @@ export async function compileServices(targetName) {
 }
 
 export async function deployServices(targetName, isForced) {
+  /**
+   * Deploying is forced by default
+   */
+  isForced = true
```
allanbowe, line 117: does this make any sense in the context of the signature in line 113?

YuryShkoda: If deploy is always forced, then we should delete `isForced` from the `cli` at all.
Instead we should update `deployServicePack` method in `adapter` to:
image

allanbowe: I agree we should remove isforced from the cli (it's redundant with this change), but not from the adapter - I think it's good to be explicit that we are going to do a full replace of the target appLoc.

medjedovicm: Makes sense

allanbowe (approved):

YuryShkoda (approved):

src/commands/deploy.js
```diff
@@ -92,7 +88,7 @@ export async function deploy(
       // split into lines
       const linesToExecute = deployScriptFile.replace(/\r\n/g, '\n').split('\n')
       if (targetToBuild.serverType === 'SASVIYA') {
-        await deployToSasViya(
+        await deployToSasAViya(
```
allanbowe, line 91: is this a typo?

src/commands/deploy.js
```diff
@@ -194,11 +190,11 @@ async function deployToSasViyaWithServicePack(buildTarget, isForced) {
       null,
       null,
       access_token,
-      isForced
+      true
     )
   }
 
-  return await sasjs.deployServicePack(jsonObject, null, null, null, isForced)
+  return await sasjs.deployServicePack(jsonObject, null, null, null)
```
saadjutt01, line 197: In this case, we are not forcing deploy? it's intentional?

saadjutt01 (dismissed):

krishna-acondy (dismissed):

PR #255, opened 2020-11-19
Issue172

### LINKED ISSUE #172: Dependencies -> SAS Macros
Currently when we perform a `sasjs compile` we check each service in the `tgtServices` folder for the existence of `  <h4> Dependencies </h4>` in the header, and then ensure that all the files below that header are included in the target service (plus a distinct list of all the recursive macro dependencies).

To clarify the intent of what we are doing, we should support the following header as well:  `  <h4> SAS Macros </h4>`

The old header should be still supported, just marked as deprecated.

- [ ] Documentation updated
- [x] Tests included (both for SAS Macros and Dependencies)
- [x] If both SAS Macros and Dependencies are provided, we use SAS Macros by default
- [x] Works both for Services, and for Jobs

medjedovicm:
## Issue

Closes #172 

## Intent

Adding support for `SAS Macros` header. `Dependencies` can still be used but it is deprecated.

## Checks

### DIFF
--- a/src/commands/build.js
+++ b/src/commands/build.js
@@ -841,7 +841,7 @@ export async function getDependencyPaths(fileContent, tgtMacros = []) {
     })
   }
 
-  const dependenciesHeader = fileContent.includes('<h4> Dependencies </h4>')
+  const dependenciesHeader = fileContent.includes('<h4> SAS Macros </h4>')
     ? '<h4> SAS Macros </h4>'
     : '<h4> Dependencies </h4>'
 
--- a/src/commands/build.js
+++ b/src/commands/build.js
@@ -825,10 +825,14 @@ export async function getDependencyPaths(fileContent, tgtMacros = []) {
       sourcePaths.push(tgtMacroPath)
     })
   }
-  const dependencies = getList(
-    '<h4> Dependencies </h4>',
-    fileContent
-  ).filter((d) => d.endsWith('.sas'))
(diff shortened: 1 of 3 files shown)


allanbowe (dismissed):

test/commands/compile/loadDependencies.spec.js
```diff
@@ -46,6 +75,7 @@ describe('loadDependencies', () => {
       [],
       'service'
     )
+    console.log(dependencies)
```
saadjutt01, line 78: Need to remove console.logs

src/commands/build.js
```diff
@@ -824,10 +824,14 @@ export async function getDependencyPaths(fileContent, tgtMacros = []) {
       sourcePaths.push(tgtMacroPath)
     })
   }
-  const dependencies = getList(
-    '<h4> Dependencies </h4>',
-    fileContent
-  ).filter((d) => d.endsWith('.sas'))
+
+  const dependenciesHeader = fileContent.includes('<h4> SAS Macros </h4>')
```
krishna-acondy, line 828: Should we display some sort of deprecation warning if the header is listed as `Dependencies` instead of `SAS Macros`?

allanbowe (approved):

allanbowe (approved):

krishna-acondy (approved):

PR #258, opened 2020-11-19
fix(job flags): --log, --output and new --status

### LINKED ISSUE #250: update syntax for 'sasjs job' command
Flag `--log` has to be changed to `--logFile` with aliases `--log` and `-l`

Flag `--output` should be changed so that it outputs to a FILE (not folder)

Flag `--status` should be added, to contain the status (eg error, success) of the job.  It should point to a FILE, and the file will contain the value of the status.  The status should also be printed to the log (success green, error red) if `--wait` is in effect.

To clarify - the final command might look like this:

```
sasjs job execute [somejob] --log /folder/file.log --output /folder/output.json --status /folder/file.txt
```

Also - we need to make sure that logs are generated when a job errors.  The tests should include failing jobs.

A job can be made to fail by including the following sas code:  `%abort;`

saadjutt01:
## Issue

Closes #250 

## Intent

To update `--log` flag, `--output` flag's behavior and add new `--status` flag
++ handle absolute paths for `output`, `logfile` and `status`
++ handle `~` for win paths
++ use default context if not specified

## Implementation

Updated command utility:
- for `--log` flag updated to `--logFile` and alias are working i.e. `--log` and `-l`
- to have another flag `--status`
Update job execute code to have `status: success` in output file in case of `wait` and print on screen.

## Checks

tested with unreleased adapter 
image

## Steps to test
- checkout branch `issue250`
- `npm install`
- `npm link`
- verify with `npm -v` it should give you the path of local repository of sasjs-cli
_You are using a linked version of SASjs CLI running from sources at 1.0.0   {PATH}_

### DIFF
--- a/src/commands/job/execute.js
+++ b/src/commands/job/execute.js
@@ -27,7 +27,7 @@ export async function execute(
   logFile,
   statusFile
 ) {
-  let result = {}
+  let result
 
   const startTime = new Date().getTime()
 
@@ -52,7 +52,12 @@ export async function execute(
     )
     .catch((err) => {
       spinner.stop()
-      if (statusFile !== undefined) displayStatus(null, statusFile, err)
+      err =
+        typeof err === 'object' && Object.keys(err).length
+          ? JSON.stringify(err)
+          : err
+      if (statusFile !== undefined)
+        displayStatus({ state: 'Error' }, statusFile, err)
       throw new Error(err)
     })
 
@@ -76,6 +81,7 @@ export async function execute(
         ? `Job located at '${jobPath}' has been executed.\nJob details`
         : `Job session`) + ` can be found at ${target.serverUrl + sessionLink}`
     )
+
     if (output !== undefined || logFile !== undefined) {
       try {
         const outputJson = JSON.stringify(submittedJob, null, 2)
@@ -165,7 +171,7 @@ async function displayStatus(submittedJob, statusFile, error = '') {
     submittedJob && submittedJob.state ? submittedJob.state : 'Not Available'
 
   const status =
-    adapterStatus === 'Not Available'
+    adapterStatus === 'Not Available' || adapterStatus === 'Error'
       ? `Job Status: ${adapterStatus}\nDetails: ${error}\n`
       : `Job Status: ${adapterStatus}`
 
--- a/src/commands/job/execute.js
+++ b/src/commands/job/execute.js
@@ -32,7 +32,7 @@ export async function execute(
   const startTime = new Date().getTime()
 
   if (statusFile !== undefined)
-    displayStatus({ state: 'Initiating' }, statusFile)
+    await displayStatus({ state: 'Initiating' }, statusFile)
 
   const spinner = ora(
     `Job located at ${chalk.greenBright(
@@ -69,7 +69,7 @@ export async function execute(
   if (result)
     displayResult(result, 'An error has occurred when executing a job.', null)
   if (statusFile !== undefined)
-    displayStatus(submittedJob, statusFile, result, true)
+    await displayStatus(submittedJob, statusFile, result, true)
 
   if (submittedJob && submittedJob.links) {
     if (!result) result = true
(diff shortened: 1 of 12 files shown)


src/commands/job/execute.js
```diff
@@ -47,14 +52,17 @@ export async function execute(
     )
     .catch((err) => {
       result = err
-
+      // need to fix this error return, it's not printing `err` on console
```
krishna-acondy, line 55: Should we not just throw an error here? It'll be caught by the layer above and be displayed in the console.

src/commands/job/execute.js
```diff
@@ -156,3 +165,27 @@ export async function execute(
...
+    submittedJob && submittedJob.state ? submittedJob.state : 'Not Available'
+  const status =
+    adapterStatus === 'Not Available'
+      ? `Job Status: ${adapterStatus}\n\nDetails: ${result}`
+      : `Job Status: ${adapterStatus}`
+
+  if (adapterStatus === 'completed') displayResult(null, null, status)
+  else displayResult(result, status, null)
+
+  if (typeof statusOfJob === 'string') {
```
krishna-acondy, line 180: The `JSDoc` comment above indicates that `statusOfJob` is a boolean.
Are we 'reusing' that to pass in a path? I also feel like the variable could be named differently to better reflect its intent.

saadjutt01: JSDoc needs to be updated, since approach is changed
Yes it's not boolean flag anymore.

src/commands/job/execute.js
saadjutt01: updated

src/commands/job/execute.js
```diff
@@ -15,6 +15,7 @@ import path from 'path'
  * @param {boolean} output - flag indicating if CLI should print out job output. If string was provided, it will be treated as file path to store output.
  * @param {boolean | string} output - flag indicating if CLI should print out job output. If string was provided, it will be treated as file path to store output. If filepath wasn't provided, output.json file will be created in current folder.
  * @param {boolean | string} log - flag indicating if CLI should fetch and save log to the local folder. If filepath wasn't provided, {job}.log file will be created in current folder.
```
allanbowe, line 17: This description should be updated (and maybe the parameter name)

For executing jobs, we always provide a `logFile` (path to a file), never a folder

test/commands/job.spec.js
```diff
@@ -98,6 +98,47 @@ describe('sasjs job', () => {
...
+        await expect(folderExists(folderPath)).resolves.toEqual(true)
+        await expect(fileExists(filePath)).resolves.toEqual(true)
+      },
+      60 * 1000
+    )
+
+    it(
+      'should submit a job and create a file with job output, log and auto-wait',
+      async () => {
+        const command = `job execute /Public/app/cli-tests/testJob -t ${targetName} -o testOutput -l testLog`
```
allanbowe, line 120: `-l testLog/some.file` should be the paramter

test/utils/command.spec.js
```diff
@@ -486,16 +486,23 @@ describe('parseCommandLine', () => {
 
     test('short syntax', () => {
       const commandLine =
-        'job execute /Public/job -t targetName -w -o ./output.json -l ./log.json'
+        'job execute /Public/job -t targetName -w -o ./output.json -l ./log.json -s ./status.txt'
```
allanbowe, line 489: the log should not arrive in JSON format (not sure if that fix is merged yet though)

allanbowe (changes requested):

allanbowe:
So the `/state` endpoint always returns successful, even when the job errors or has warnings.

We should query the `stateDetails` instead, in order to find out the state of the job.  

https://developer.sas.com/apis/rest/v3.5/Compute/#job-39-s-statedetails-registry

allanbowe (approved):

krishna-acondy (approved):

PR #353, opened 2020-12-15
fix(create): fetching template info without curl

### LINKED ISSUE #321: certificate issue in portable git bash on windows
Problem:

The `sasjs create someProject -t jobs` command fails with an exception

Cause:

When running portable git bash on a locked down windows machine the underlying `curl` command fails with a certificate error:

image

Interestingly the same command works fine in the `cmd` shell.  

Action:
Rather than throw an exception we should handle this case and present the result back to the user.

saadjutt01:
## Issue

Closes #321 

## Intent

Remove `curl`'s only existence from code, it requires certificates to be configured. Some users have issue on windows with portable git-bash

## Implementation

removed `curl` (calling github api), replaced with `git ls-remote`

## Checks

- [x] Reviewer is assigned.
image

To test:
checkout branch `issue321` and `npm start`

### DIFF
--- a/src/utils/utils.js
+++ b/src/utils/utils.js
@@ -38,18 +38,14 @@ export async function createMinimalApp(folderPath) {
 
 export async function createTemplateApp(folderPath, template) {
   const { stdout, stderr, code } = shelljs.exec(
-    `git ls-remote git@github.com:sasjs/template_${template}`,
+    `git ls-remote https://username:password@github.com/sasjs/template_${template}.git`,
     {
       silent: true
     }
   )
 
-  if (stderr.startsWith('ERROR: Repository not found.')) {
+  if (stderr.includes('Repository not found') || code) {
     throw `Template "${template}" is not sasjs template`
-  }
-
-  if (code) {
-    throw `\n${stderr}`
   }
 
   if (!stdout) {
--- a/src/utils/utils.js
+++ b/src/utils/utils.js
@@ -37,19 +37,24 @@ export async function createMinimalApp(folderPath) {
 }
 
 export async function createTemplateApp(folderPath, template) {
-  const {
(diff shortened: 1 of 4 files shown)


src/utils/utils.js
```diff
@@ -37,19 +37,24 @@ export async function createMinimalApp(folderPath) {
 }
 
 export async function createTemplateApp(folderPath, template) {
-  const {
-    stdout
-  } = shelljs.exec(
-    `curl https://api.github.com/repos/sasjs/template_${template}`,
-    { silent: true }
+  const { stdout, stderr, code } = shelljs.exec(
+    `git ls-remote git@github.com:sasjs/template_${template}`,
```
allanbowe, line 41: Will this approach (git@github) still work even if the user does not have an ssh key?

saadjutt01: yes, because our template repositories are public. No ssh key is required.

allanbowe (approved):

allanbowe (dismissed):

src/utils/utils.js
```diff
@@ -37,19 +37,20 @@ export async function createMinimalApp(folderPath) {
 }
 
 export async function createTemplateApp(folderPath, template) {
-  const {
-    stdout
-  } = shelljs.exec(
-    `curl https://api.github.com/repos/sasjs/template_${template}`,
-    { silent: true }
+  const { stdout, stderr, code } = shelljs.exec(
+    `git ls-remote https://username:password@github.com/sasjs/template_${template}.git`,
```
allanbowe, line 41: so the dummy username:password prevents the authentication prompt when the user provides the wrong template name?

Nice!

allanbowe (approved):

Issue #302, opened 2020-12-04
Refactor code to use new Target model


krishna-acondy:
We would like to use these models for the target going forward.

```
import { ServerType } from './serverType'

export interface Config {
  macroVars: { [key: string]: string }
  initProgram: string
  termProgram: string
}

export interface BuildConfig extends Config {
  buildOutputFileName: string
}

export interface DeployConfig {
  deployServicePack: boolean
  deployScripts: string[]
}

export interface ServiceConfig extends Config {
  serviceFolders: string[]
}

export interface JobConfig extends Config {
  jobFolders: string[]
}

export interface StreamConfig {
  assetPaths: string[]
  streamWeb: boolean
  streamWebFolder: string
  webSourcePath: string
}

export interface Target {
  name: string
  serverUrl: string
  serverType: ServerType
  appLoc: string
  buildConfig: BuildConfig
  deployConfig: DeployConfig
  //   deployVars: { [key: string]: string }
  serviceConfig: ServiceConfig
  jobConfig: JobConfig
  streamConfig: StreamConfig
  macroFolders: string[]
  programFolders: string[]
}
```

Issue #269, opened 2020-11-24
mark `tgtDeployVars` as deprecated, replace with `serviceVars`


allanbowe:
We should add a one year deployment warning on targets that contain `xxxDeployVars` and replace with `serviceVars` (without cmn and tgt prefix)

The purpose of these name/value pairs is to add them to the start of each SAS _service_ as SAS Macro variables.  

If a config file contains both "common" and "target" `serviceVars` then any same-named "target" vars _replace_ same-named "common" vars.  The result may contain (differently named) vars from _both_ common & target.

**The `tgtDeployVars` object should not be used by the adapter, or the CLI, internally**.  If the tests / CLI / adapter are using the contextName from this location, then they should be refactored to take the contextName from the root of the target definition instead.  If no contextname is found, the default should be "SAS Job Execution compute context".

(Historical context - they were originally put there deliberately to avoid duplication as we were using CLIENT & SECRET in both frontend and backend.  Not any more.)

Issue #185, opened 2020-10-29
Centralise logging and provide different log levels


krishna-acondy:
At the moment, we are using `chalk` all over the code to show console messages in different ways.

It'd be good to centralise logging into a utility class or set of functions that wraps around `chalk` and provides methods like `warn`, `debug`, `error` and `info`. We can then control what is shown on screen by setting the log level (maybe in the config, or as an environment variable).

Issue #165, opened 2020-10-14
deprecate cmn and tgt prefixes


allanbowe:
To avoid breaking changes, we should keep the old variables functional, until we do eventually bump to v2.0

To clarify, the changes are such:

* cmnServices -> serviceFolders
* tgtServiceInit -> serviceInitProgram
* cmnMacros -> macroFolders

etc

PR #357, opened 2020-12-17
BREAKING CHANGE: SASjs CLI 2.0 - Electric Boogaloo

krishna-acondy:
BREAKING CHANGE: SASjs CLI 2.0 Release

## Issue

Fixes https://github.com/sasjs/cli/issues/269
Fixes https://github.com/sasjs/cli/issues/185
Fixes https://github.com/sasjs/cli/issues/165
Fixes https://github.com/sasjs/cli/issues/287
Fixes https://github.com/sasjs/cli/issues/302
Fixes https://github.com/sasjs/cli/issues/358

## Intent

* Rewrite the CLI to improve stability, reduce bugs and enhance developer experience.
* Make the implementation work with the new `Target` model defined in `@sasjs/utils`.
* Improve the stability of the unit tests.
* Separate the server-based tests from mocked ones.

### Targets
* There is now ONE single way of retrieving a target - `findTargetInConfiguration` in `src/utils/config`.
* It returns an object containing a `target` which is of type `Target` from `@sasjs/utils`, and a boolean `isLocal` which indicates whether the target is from the local config or the global RC file.
* `Target`s are read-only. That means the only way to modify them is by changing the values in the JSON files, or by replacing existing targets with new ones.
* When a target is saved to a config file, we call the `target.toJson()` method on it, which will convert it into a JSON representation suitable for storage.
* When retrieving a target, the JSON is used to create a `Target` object.

### Project Architecture
* Each command has its own folder within `src/commands`.
* Inside that folder, there will be one module that will export the main function(s) for that command.
* There is a subfolder within that folder called `internal`, which will contain any helper functions specific to this command.
* We should strive to split functionality out into multiple modules so we don't end up with 1000+ line files.
* Splitting out functions also gives us the ability to mock out entire modules using Jest when testing our code.

### TypeScript
* All code and tests have now been migrated over to TypeScript
* New interfaces and types have been added where applicable.

### Error Handling
* The individual command implementations will throw any errors that they encounter.
* The error will be handled in `main.ts`.
* The functions in `main.ts` will return a `ReturnCode` to the CLI, which will be used as the exit code.
* The different `ReturnCode`s available are `Success`, `InvalidCommand` and `InternalError`.
* We will be able to use this mechanism to show friendly messages to make things clearer to the user.
* There is still some iffy code in some of the commands which we need to refactor.
* All command implementations should be throwing errors as soon as possible in the process - failing fast is important and will help guard against getting into inconsistent states.

### Logging
* A `Logger` (from `@sasjs/utils`) is now instantiated at the very beginning when any command is run.
* The logger is available on the `process` object as `process.logger`.
* The log level is based on a `LOG_LEVEL` env variable and defaults to `Error` (might make sense to default to `Debug` or `Info` in the CLI).
* Most instances of using `console.log` along with `chalk` have been refactored to use the new logger.
* `sasjs help` is still using `chalk` since it is actually required there to make the text more easily readable.

### Testing
* Most of the functions from `helpers.js` which were being attached to the `global` object have been removed.
* Some have been moved into the `utils/test` module. This module and all associated files are excluded from the built package.
* The separate `test` folder has been removed, and the tests for each command are in a `spec` folder alongside the actual command.
* All the tests have been refactored to create their test projects within the current folder rather than at the root level.
* This will reduce the likelihood of clashes between different test suites.
* All test suites also use unique timestamp-based names for their targets and apps.
* The `spec` files are also excluded from the final bundle.
* There are utilities added to `utils/test` for creating and tearing down test apps and targets.
* There are now two separate `jest` configurations - one for mocked/non-server tests, and one for server-based tests.
* When creating server-based tests, you will need to create files called `<name>.spec.server.ts`.
* When creating non-server tests, you will need to create files called `<name>.spec.ts`.
* The CI workflows have been updated to run the 'smoke tests' first.
* Then, they'll run all the non-server tests - this takes less than a minute.
* If all of them pass, they'll run the server-based tests, which may take upto 10 minutes.

### Compile / Build / Deploy
* I've separated these into their own individual commands to split out the code better and make them independent.
* The compilation check is now done in the `compile` module - so we won't recompile if not required.
* The `build` command will call compile, which will recompile if necessary.
* When running `cb` or `cbd` commands, there is now error handling at each step of the process - so if something fails on the `compile` step, we won't attempt to execute any further steps and so on.

phew

## Checks

- [x] Reviewer is assigned.


package.json
```diff
@@ -42,7 +44,7 @@
   "dependencies": {
     "@sasjs/adapter": "^1.19.4",
     "@sasjs/core": "^1.12.3",
```
allanbowe, line 46: This can go to v2 now

src/cli.spec.ts
```diff
@@ -0,0 +1,387 @@
...
+      appName,
+      `/Public/app/cli-tests/${appName}`
+    )
+    done()
+  })
+
+  afterEach(async (done) => {
+    await removeTestApp(__dirname, target.name)
+    await removeFromGlobalConfig(target.name)
+    jest.clearAllMocks()
```
YuryShkoda, line 30: Do we still need to clear it here (we've set `clearMocks: true` in jest config)?

src/commands/compile/internal/config.ts
```diff
@@ -0,0 +1,115 @@
...
+      configuration.jobConfig &&
+      configuration.jobConfig.initProgram
+    ) {
+      jobInitContent = await readFile(
+        path.join(buildSourceFolder, configuration.jobConfig.initProgram)
+      )
+    }
+  }
+
+  console.log('JOBINIT: ', jobInitContent)
```
YuryShkoda, line 83: We don't need this `console.log()`

src/utils/config.ts
```diff
@@ -1,26 +1,29 @@
...
-    Promise.resolve(null)
-  })
+/**
+ * Returns an object that represents the SASjs CLI configuration in a given file.
+ * @param {string} pathToFile - the path to the file in question.
+ * @returns {Configuration} configuration object if available.
+ */
+export async function getConfiguration(
+  pathToFile: string
+): Promise<Configuration> {
```
YuryShkoda, line 18: this function doesn't return a Promise

src/utils/config.ts
```diff
@@ -30,70 +33,73 @@ export async function getConfiguration(pathToFile) {
...
  * @param {boolean} viyaSpecific - will fall back to the first target of type SASVIYA.
+ * @returns {Target} target or fallback when one is found.
  */
 export async function findTargetInConfiguration(
-  targetName,
+  targetName: string,
   viyaSpecific = false
-) {
-  const projectRoot = await getProjectRoot()
+): Promise<{ target: Target; isLocal: boolean }> {
```
YuryShkoda, line 41: this function doesn't return a Promise

YuryShkoda (dismissed):

package.json
krishna-acondy: Done!

src/commands/compile/internal/config.ts
krishna-acondy: Removed.

src/utils/config.ts
krishna-acondy: `async/await` in ES6/TypeScript is based on `Promise`s, so this function _does_ actually return a `Promise`.

src/utils/config.ts
krishna-acondy: `async/await` in ES6/TypeScript is based on `Promise`s, so this function _does_ actually return a `Promise`.

medjedovicm (approved):

saadjutt01 (dismissed):

src/utils/config.ts
```diff
@@ -0,0 +1,464 @@
...
+    path.join(projectRoot, 'sasjs', 'sasjsconfig.json')
+  ).catch(() => null)
+  if (localConfig && localConfig.programFolders) {
+    programFolders = programFolders.concat(localConfig.programFolders)
+  }
+
+  const { target } = await findTargetInConfiguration(targetName)
+
+  if (!target) {
+    throw new Error(
```
saadjutt01, line 263: this piece of code is unreachable (and redundant as code throws in `findTargetInConfiguration` if there's no target.

src/utils/config.ts
```diff
@@ -0,0 +1,464 @@
...
+    path.join(projectRoot, 'sasjs', 'sasjsconfig.json')
+  ).catch(() => null)
+  if (localConfig && localConfig.programFolders) {
+    macroFolders = macroFolders.concat(localConfig.programFolders)
+  }
+
+  const { target } = await findTargetInConfiguration(targetName)
+
+  if (!target) {
+    throw new Error(
```
saadjutt01, line 300: this piece of code is unreachable (and redundant as code throws in `findTargetInConfiguration` if there's no target.

saadjutt01 (changes requested):

src/utils/config.ts
krishna-acondy: I've removed that bit of code.

src/utils/config.ts
krishna-acondy: I've removed that bit of code.

src/cli.spec.ts
krishna-acondy: Keeping it there just in case.

saadjutt01 (approved):

PR #383, opened 2021-01-05
Issue 174

### LINKED ISSUE #174: sasjs folder move
There is a bug in `sasjs folder move`

image

The result was:

image

In any case, the behaviour may need to be revisited.  It would be more useful / intuitive if the move command could move AND rename a folder, in which case it would be more like the linux `mv`.  The behaviour would depend on the presence of a trailing `/`.  If a trailing `/` is provided, the folder is moved to the specified parent.  If none provided, then the folder is moved and renamed.

`sasjs folder move <sourcefolder> <targetfolder>`

To enable automated testing, we should also create a `sasjs folder list` command, which will list all the immediate children of a folder (and their types).

## Definition of done:

- [x]  `sasjs folder move /Public/example1 /Public/example2` will rename example1 to example2 (unless example2 exists, in which case it will fail)
- [x] `sasjs folder move /Public/nested/folder /Public/` will move `folder` to `/Public/folder`
- [x]  the presence of a _leading_ `/` determines whether to prepend the path with the value of `appLoc`
- [ ] `sasjs folder list` command created & documented
- [x]  we have automated tests to confirm the above behaviours

medjedovicm:
## Issue

Closes #174 

## Intent

Fixing folder `move` command and adding `list` command. Also adding server tests for each operation.

## Checks

- [x] Reviewer is assigned.


src/commands/folder/spec/folder.spec.server.ts
```diff
@@ -0,0 +1,191 @@
...
+import { folder } from '../index'
+import { generateTimestamp } from '../../../utils/utils'
+import { ServerType, Target } from '@sasjs/utils/types'
+import {
+  removeFromGlobalConfig,
+  saveToGlobalConfig
+} from '../../../utils/config'
+import { Command } from '../../../utils/command'
+
+describe('sasjs job execute', () => {
```
YuryShkoda, line 11: should describe `sasjs folder` command

src/commands/folder/spec/folder.spec.server.ts
```diff
@@ -0,0 +1,191 @@
...
+    done()
+  })
+
+  it('list folder children', async (done) => {
+    const timestamp = generateTimestamp()
+    const testFolderPath = `/Public/app/cli-tests/cli-tests-folder-${timestamp}`
+
+    await createTestFolder(testFolderPath, target.name)
+
+    const commandList1 = new Command(
```
YuryShkoda, line 32: there is no need to add `1` to the variable name here

src/commands/folder/spec/folder.spec.server.ts
```diff
@@ -0,0 +1,191 @@
...
+      `folder list /Public/app/cli-tests -t ${target.name}`
+    )
+    const list1 = await folder(commandList1)
```
YuryShkoda, line 35: there is no need to add `1` to the variable name here

src/commands/folder/spec/folder.spec.server.ts
```diff
@@ -0,0 +1,191 @@
...
+    expect(list1).toContain(`cli-tests-folder-${timestamp}`)
```
YuryShkoda, line 36: it would be better to use RegExp to verify that `cli-tests-folder-${timestamp}` entry is in the right place

src/commands/folder/spec/folder.spec.server.ts
```diff
@@ -0,0 +1,191 @@
...
+  })
+
+  afterAll(async (done) => {
+    await removeFromGlobalConfig(target.name)
+    done()
+  })
+
+  it('list folder children', async (done) => {
+    const timestamp = generateTimestamp()
+    const testFolderPath = `/Public/app/cli-tests/cli-tests-folder-${timestamp}`
```
YuryShkoda, line 28: `testFolderPath` is the same for every test suit, but with a different `timestamp`, so you can create a function returning `testFolderPath`

src/commands/folder/spec/folder.spec.server.ts
```diff
@@ -0,0 +1,191 @@
...
+    done()
+  })
+
+  it('move folders keeping the folder name', async (done) => {
+    const timestamp = generateTimestamp()
+    const testFolderPath = `/Public/app/cli-tests/cli-tests-folder-${timestamp}`
+
+    await createTestFolder(testFolderPath, target.name)
+
+    const commandList1 = new Command(
```
YuryShkoda, line 48: there is no need to add 1 to the variable name here

src/commands/folder/spec/folder.spec.server.ts
```diff
@@ -0,0 +1,191 @@
...
+      `folder list /Public/app/cli-tests -t ${target.name}`
+    )
+    const list1 = await folder(commandList1)
```
YuryShkoda, line 51: there is no need to add 1 to the variable name here

src/commands/folder/spec/folder.spec.server.ts
```diff
@@ -0,0 +1,191 @@
...
+      `folder create ${testFolderPath}/temp/test -t ${target.name}`
+    )
+    await folder(commandCreate2)
+
+    const commandCreate3 = new Command(
+      `folder create ${testFolderPath}/test2 -t ${target.name}`
+    )
+    await folder(commandCreate3)
+
+    const commandMove1 = new Command(
```
YuryShkoda, line 64: there is no need to add 1 to the variable name here

src/commands/folder/spec/folder.spec.server.ts
```diff
@@ -0,0 +1,191 @@
...
+      `folder move ${testFolderPath}/temp ${testFolderPath}/test2 -t ${target.name}`
+    )
+    await folder(commandMove1)
+
+    // Check if operations are executed correctly
+    let folderList1 = await folder(
+      new Command(`folder list ${testFolderPath}/test2 -t ${target.name}`)
+    )
+    expect(folderList1).toContain('temp')
```
YuryShkoda, line 73: it would be better to use RegExp

src/commands/folder/spec/folder.spec.server.ts
```diff
@@ -0,0 +1,191 @@
...
+
+    let folderList2 = await folder(
+      new Command(`folder list ${testFolderPath}/test2/temp -t ${target.name}`)
+    )
+    expect(folderList2).toContain('test')
```
YuryShkoda, line 78: it would be better to use RegExp

src/commands/folder/spec/folder.spec.server.ts
```diff
@@ -0,0 +1,191 @@
...
+
+    await deleteTestFolder(testFolderPath, target.name)
+    done()
+  })
+
+  it('move folder to same location renaming the folder', async (done) => {
```
YuryShkoda, line 84: `move folder to same location renaming the folder` -> `move folder to the same location and rename it`

YuryShkoda (changes requested):

YuryShkoda (approved):

PR #430, opened 2021-01-26
feat(add): make insecure request

### LINKED ISSUE #428: Add  enable insecure connection to add cred command
**Syntax**: `sasjs add cred -t target --insecure` 

**Issue example**: 
image

**Definition of done**: user should be able to run `sasjs add cred` to run in insecure mode against servers with self signed certificates.  When insecure mode is enabled, we should have a warning in the log.

If `--insecure` flag was provided, CLI should display a warning.

saadjutt01:
## Issue

Closes #428 

## Intent

Server response with _self signed certificate_ should be accepted if `--insecure` flag is present

## Implementation

Passed `insecure` boolean to adapter

## Checks

- [ ] Reviewer is assigned.

Cli: mocked tests
image

Cli: server tests
image

### DIFF
--- a/src/commands/add/addCredential.ts
+++ b/src/commands/add/addCredential.ts
@@ -17,8 +17,13 @@ import { getAndValidateServerUrl, getCredentialsInput } from './internal/input'
  * The file will contain the client ID, secret, access token and refresh token.
  * Its name will be of the form `.env.{targetName}`
  * @param {string} targetName- the name of the target to create the env file for.
+ * @param {boolean} insecure- boolean to use insecure connection, default is false.
  */
-export const addCredential = async (targetName: string): Promise<void> => {
+export const addCredential = async (
+  targetName: string,
+  insecure: boolean = false
+): Promise<void> => {
+  if (insecure) process.logger?.warn('Executing with insecure connection.')
   targetName = validateTargetName(targetName)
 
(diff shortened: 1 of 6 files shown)


src/commands/add/addCredential.ts
```diff
@@ -17,8 +17,13 @@ import { getAndValidateServerUrl, getCredentialsInput } from './internal/input'
  * The file will contain the client ID, secret, access token and refresh token.
  * Its name will be of the form `.env.{targetName}`
  * @param {string} targetName- the name of the target to create the env file for.
+ * @param {boolean} insecure- boolean to use insecure connection, default is false.
```
allanbowe, line 20: As a reader I would be curious what an insecure connection means, and why we would need it. Can we provide a link to further information?  Eg the adapter code, or documentation, or the updated page on cli.sasjs.io ?

allanbowe: or, (or in addition) could of course provide more information here.

YuryShkoda (approved):

src/commands/add/addCredential.ts
saadjutt01: Updating to: 
* @param {boolean} insecure- boolean to use insecure connection, default is false. If true the cli will not reject server connection which is not authorized with the list of supplied CAs.

YuryShkoda (approved):

PR #446, opened 2021-02-05
fix(add): parsing target with target.json()

### LINKED ISSUE #433: issue adding url with sasjs add cred
We have an issue with SASjs add cred when inserting the url

https://asciinema.org/a/C9cNZBYxn0QqOh7EmDbJsqwcO

steps to reproduce:

```
sasjs create test -t jobs
sasjs add cred -t viya
```
then add https://sas.analytium.co.uk

The cause could be that the serverUrl is not in the sasjs config, if so we should build a test for this case.  It could also be special chars from the prompt.

saadjutt01:
## Issue

Closes #433 

## Intent

Adding server url to target is breaking

## Implementation

parsed `target` with `target.json()`

## Checks

- [x] Reviewer is assigned.

### DIFF
--- a/src/commands/add/spec/add-credential.spec.ts
+++ b/src/commands/add/spec/add-credential.spec.ts
@@ -21,7 +21,7 @@ describe('addCredential', () => {
 
     expect(inputModule.getAndValidateServerUrl).toHaveBeenCalled()
     expect(configUtils.saveToLocalConfig).toHaveBeenCalledWith(
-      new Target({
+      jasmine.objectContaining({
         name: 'test-target',
         serverUrl: 'http://server.com',
         serverType: ServerType.SasViya,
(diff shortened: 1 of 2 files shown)


src/commands/add/spec/add-credential.spec.ts
```diff
@@ -21,7 +21,7 @@ describe('addCredential', () => {
 
     expect(inputModule.getAndValidateServerUrl).toHaveBeenCalled()
     expect(configUtils.saveToLocalConfig).toHaveBeenCalledWith(
-      new Target({
+      jasmine.objectContaining({
```
YuryShkoda, line 24: why do we need `jasmine` here?

saadjutt01: `Target` has all the attributes with default values, to partial check with concerned attributes.

https://jasmine.github.io/2.0/introduction#section-Partial_Matching_with_%3Ccode%3Ejasmine.objectContaining%3C/code%3E

krishna-acondy: This is the `jest` way of doing it: https://jestjs.io/docs/en/expect#expectobjectcontainingobject

krishna-acondy (approved):

PR #448, opened 2021-02-06
feat(lineage): added lineage to 'sasjs doc'

### LINKED ISSUE #425: Data Lineage diagram
Using graphviz (which I think may come bundled inside Doxygen) we can generate data lineage diagrams.

These should be based on the following entries in SAS Jobs and Services:

```
  <h4> Data Inputs </h4>
  @li LIBREF.TABLENAME
  @li lib2.othertable

  <h4> Data Outputs </h4>
  @li somelib.table4
 ```

We should scan the jobFolders and serviceFolders for such files (perhaps it can be configurable in the docConfig, the list of types to scan, eg jobs/services/macros/programs)

The generated graphviz diagram should link the programs to the relevant program in the doxygen generated documentatoin.

The tables should link to Data Controller, if a `datacontroller_url` is available in the docConfig.

The diagram should visible on a new page on the generated doxygen documentation (I think we need to re-enable the 'tabs' as can be seen on the doxygen homepage)

### LINKED ISSUE #449: sasjs doc should only pick up one target
Previously, when running `sasjs doc` without a target, we would document all the SAS macros / programs / jobs / services from all targets.

Now that we are introducing [data controller](https://datacontroller.io) as an attribute (`dataControllerUrl`) this approach no longer makes sense.  

When building documentation for a target (eg DEV or TEST), now that we link the tables to the data controller viewer, we can only reference one target.

So moving forward, if no target is supplied, we should choose the first target (like we do elsewhere in the CLI).  If a target is supplied, we use that target (like we do now).   If a `docConfig` object exists in a target, it's attributes take precedence over the root `docConfig`.

### PR DESCRIPTION
**Pull request by saadjutt01**
## Issue

Closes #425 
Closes #449 

## Intent

Add `sasjs doc lineage`
Create dot file
Use dot to create diagram
Render inside doxygen documentation

Update DocConfig and also added to Target 

## Implementation

## Checks

- [x] Reviewer is assigned.

### CODE REVIEW

File: src/commands/docs/generateDot.ts [hunk 1]
```diff
@@ -0,0 +1,40 @@
...
+import { getLocalConfig } from '../../utils/config'
+import { getFoldersForDocs } from './internal/getFoldersForDocs'
+import { createDotFiles } from './internal/createDotFiles'
+import { getDocConfig } from './internal/getDocConfig'
+
+/**
+ * Generates lineage in dot language
+ * By default the dot files will be at 'sasjsbuild/docs' folder
+ * If a target is supplied, generates dot files only for the jobs / services in that target (and the root).
+ * If no target is supplied, generates for all jobs / services.
```
allanbowe, line 12: I think this comment is not valid any more

saadjutt01: Yes, I need to update/add comments and tests for `sasjs doc lineage`

src/commands/docs/internal/createDotFiles.ts
```diff
@@ -0,0 +1,25 @@
...
+  const dotFilePath = path.join(outDirectory, 'generated_code.dot')
+  const dotGraphPath = path.join(outDirectory, 'graph_diagram.svg')
+
+  await createFolder(outDirectory)
+
+  const dotFileContent = await getDotFileContent(folderList, serverUrl)
+
+  await createFile(dotFilePath, dotFileContent)
+
+  const { graphviz } = require('node-graphviz')
```
krishna-acondy, line 21: Could we please use an `import` instead of `require` like we do for everything else?

saadjutt01: done

src/commands/docs/generateDocs.ts
```diff
@@ -5,41 +5,52 @@ import ora from 'ora'
...
+import { getDocConfig } from './internal/getDocConfig'
 
 /**
  * Generates documentation(Doxygen)
  * By default the docs will be at 'sasjsbuild/docs' folder
  * If a target is supplied, generates docs only for the SAS programs / macros / jobs / services in that target (and the root).
  * If no target is supplied, generates for all sas programs/ macros / jobs / services.
  * @param {string} targetName- the name of the target to be specific for docs.
- * @param {string} outDirectory- the name of the output folder, picks from sasjsconfig.docConfig if present.
+ * @param {string} outDirectoryInput- the name of the output folder, picks from sasjsconfig.docConfig if present.
```
medjedovicm, line 21: Why is in `outDirectoryInput` word `input` more descriptive?

krishna-acondy: I agree. I think outDirectory was more descriptive than outDirectoryInput.

saadjutt01: Updated

allanbowe (approved):

PR #457, opened 2021-02-09
fix: removed flag --insecure and now using Target.allowInsecureRequests

### LINKED ISSUE #455: Allow insecure request via Target property and remove --insecure flag
Currently implemented, 
https://github.com/sasjs/cli/issues/428

it's for only `sasjs add cred`, but it's required for all commands, hence need to remove `--insecure` flag and add boolean to `sasjsconfig.json`

Add Target property "allowInsecureRequests", defaults to false.

saadjutt01:
## Issue

Closes #455 

## Intent

Instead of flag from commandline, config `allowInsecureRequests` is used, in root as well as at Target level.
Flag `--instead` will only be available for `sasjs add`

## Implementation

`--insecure` flag captured for `sasjs add target` and `sasjs add cred`, having precedence over config.

`allowInsecureRequests` extracted from target & config, then provided to all occurrences of SASjs instance,
- `findTargetInConfiguration` is updated having `allowInsecureRequests` attribute based on root level.
- Having target level precedence and defaults to false, if not found at root.

## Checks

- [x] Reviewer is assigned.


allanbowe (dismissed):

src/commands/add/addCredential.ts
```diff
@@ -21,13 +21,16 @@ import { getAndValidateServerUrl, getCredentialsInput } from './internal/input'
  */
 export const addCredential = async (
   targetName: string,
-  insecure: boolean = false
+  insecure: boolean | undefined = undefined
```
krishna-acondy, line 24: Is this really necessary?

saadjutt01: Actually, no.
thanks for highlighting, I have double-check and updated.

allanbowe (approved):

krishna-acondy (approved):
Looks good to me!

PR #517, opened 2021-02-24
feat(compile): compile single job/service

### LINKED ISSUE #469: sasjs compile for a single file
For a large project it can be time consuming to compile and deploy every service.  In addition, for our upcoming VSCode plugin, we need a way to compile and run a single Job or Service.

Proposal - we add two new actions to the `sasjs compile` command, eg:

`sasjs compile job -s myjob.sas -t mytarget -o /some/folder`

and

`sasjs compile service -s myservice.sas -t mytarget`

So will then have three 'compile' commands:

* `sasjs  compile` -> compiles everything in a SASjs project, works locally
* `sasjs compile  job` -> compiles  a single job, can work anywhere
* `sasjs compile service` -> compiles a single service, can work anywhere

Arguments:

--source (alias -s) - the path/name.ext of the individual source file to compile (as job or service).  MANDATORY
--target (alias -t) - The target to use for obtaining the source folders of programs and macros.  If it is not specified, the first target is used by default.  The target can exist either in the local project configuration or in the global .sasjsrc file.  OPTIONAL
--output (alias -o) - path where output of the compiled job or service will be saved.  OPTIONAL.  If not provided, the output will go to the root of the sasjsbuild folder if in a project (sasjsbuild would be emptied first), else in the current working directory.  

The created job or service will have the same name as the source file.

### LINKED ISSUE #486: sasjs should support absolute paths
Whilst working on a temporary project I tried to add a path to the macros of a different project:

image

When running `sasjs compile` I received the following error:

image

We should support absolute paths (identified by a leading slash, or a drive letter on windows) everywhere within the project.

saadjutt01:
## Issue

Closes #469 
Closes #486 

## Intent

Compile single job/service.
And to support absolute paths wherever possible

## Implementation

Code refactored:
- extracted code from compile to internals, `compileServiceFile.ts` and `compileJobFile.ts`
- used these to newly introduced command file `compileSingleFile.ts`

Fix in context index.ts
`command.ts` utility updated accordingly to new supported flags for subcommand(`job` and `service`)

changed `path.join` to if `path.isAbsolute` use it else use `path.join`

## Checks

- [x] Reviewer is assigned.

### DIFF
--- a/src/commands/compile/compileSingleFile.ts
+++ b/src/commands/compile/compileSingleFile.ts
@@ -36,8 +36,12 @@ export async function compileSingleFile(
     throw new Error(`Provide a path to source file (eg '${commandExample}')`)
   }
 
-  const sourcePath = path.resolve(process.projectDir, source)
-  const outputPath = path.resolve(process.projectDir, output ?? './sasjsbuild')
+  const sourcePath = path.isAbsolute(source)
+    ? source
+    : path.join(process.projectDir, source)
+  const outputPath = path.isAbsolute(output)
+    ? output
+    : path.join(process.projectDir, output ?? './sasjsbuild')
 
   process.logger?.info(`Compiling source file:\n- ${sourcePath}`)
 
--- a/src/commands/compile/compileSingleFile.ts
+++ b/src/commands/compile/compileSingleFile.ts
@@ -78,7 +78,7 @@ export async function compileSingleFile(
 async function validateSourcePath(path: string) {
   if (!path) return false
 
-  const isSourceFile = /\.sas/i.test(path)
+  const isSourceFile = /\.sas$/i.test(path)
 
   if (!isSourceFile) return false
 
--- a/src/commands/compile/compileSingleFile.ts
+++ b/src/commands/compile/compileSingleFile.ts
@@ -36,8 +36,8 @@ export async function compileSingleFile(
     throw new Error(`Provide a path to source file (eg '${commandExample}')`)
   }
 
-  const sourcePath = path.join(process.cwd(), source)
-  const outputPath = output ?? './sasjsbuild'
+  const sourcePath = path.resolve(process.projectDir, source)
+  const outputPath = path.resolve(process.projectDir, output ?? './sasjsbuild')
 
   process.logger?.info(`Compiling source file:\n- ${sourcePath}`)
 
@@ -45,7 +45,7 @@ export async function compileSingleFile(
   await createFolder(outputPath)
 
   const sourceFileName = sourcePath.split(path.sep).pop() as string
-  const destinationPath = path.join(process.cwd(), outputPath, sourceFileName)
+  const destinationPath = path.join(outputPath, sourceFileName)
   await copy(sourcePath, destinationPath)
 
   const macroFolders = target ? target.macroFolders : []
@@ -61,13 +61,13 @@ export async function compileSingleFile(
       )
       break
     case subCommands.job:
-      break
       await compileJobFile(
         target,
         destinationPath,
         macroFolders,
         programFolders
       )
+      break
     default:
       break
   }
--- a/src/commands/compile/compileSingleFile.ts
+++ b/src/commands/compile/compileSingleFile.ts
@@ -1,0 +1,86 @@
+import path from 'path'
+import { getProgramFolders } from '../../utils/config'
+import { copy, fileExists, deleteFolder, createFolder } from '../../utils/file'
+import { Target } from '@sasjs/utils/types'
+import { compileServiceFile } from './internal/compileServiceFile'
+import { compileJobFile } from './internal/compileJobFile'
+import { Command } from '../../utils/command'
+
+export async function compileSingleFile(
+  target: Target,
+  command: Command,
+  subCommand: string
+) {
+  const subCommands = {
+    job: 'job',
+    service: 'service'
+  }
+
+  if (!subCommands.hasOwnProperty(subCommand)) {
+    throw new Error(
+      `Unsupported context command. Supported commands are:\n${Object.keys(
+        subCommands
+      ).join('\n')}`
+    )
+  }
+
+  const commandExample =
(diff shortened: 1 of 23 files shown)


src/commands/compile/compileSingleFile.ts
```diff
@@ -0,0 +1,86 @@
...
+  } else if (!(await validateSourcePath(source))) {
+    throw new Error(`Provide a path to source file (eg '${commandExample}')`)
+  }
+
+  const sourcePath = path.resolve(process.projectDir, source)
+  const outputPath = path.resolve(process.projectDir, output ?? './sasjsbuild')
+
+  process.logger?.info(`Compiling source file:\n- ${sourcePath}`)
+
+  await deleteFolder(outputPath)
```
YuryShkoda, line 44: in `sasjsbuild` can be some data, why do we need to delete it?

src/commands/compile/compileSingleFile.ts
```diff
@@ -0,0 +1,86 @@
...
+      break
+  }
+
+  return { destinationPath }
+}
+
+async function validateSourcePath(path: string) {
+  if (!path) return false
+
+  const isSourceFile = /\.sas/i.test(path)
```
YuryShkoda, line 81: `.sas` should be the end of the string

src/commands/compile/internal/compileServiceFile.ts
```diff
@@ -0,0 +1,69 @@
...
+    macroFolders,
+    programFolders
+  )
+
+  const serverType = await getServerType(target)
+  const preCode = await getPreCodeForServicePack(serverType)
+
+  dependencies = `${preCode}\n${dependencies}`
+
+  if (dependencies) await createFile(filePath, dependencies)
```
YuryShkoda, line 27: `dependencies` will always be true in this case

src/commands/compile/spec/compile.spec.ts
```diff
@@ -81,3 +86,63 @@ describe('sasjs compile', () => {
     done()
   })
 })
+
+describe('sasjs compile', () => {
```
YuryShkoda, line 90: we've already described `sasjs compile` at line 16

src/commands/compile/spec/compile.spec.ts
```diff
@@ -81,3 +86,63 @@ describe('sasjs compile', () => {
...
+      await expect(
+        compileSingleFile(
+          target,
+          new Command(
+            `compile job -s ${process.projectDir}/jobs/extract/makedata1.sas`
+          ),
+          'job'
+        )
+      ).toResolve()
+      expect(compileJobFile.compileJobFile).toHaveBeenCalled()
```
YuryShkoda, line 119: The fact that `compileJobFile` has been called doesn't prove that a single file has been compiled properly and saved at the correct location.

src/commands/compile/spec/compile.spec.ts
```diff
@@ -81,3 +86,63 @@ describe('sasjs compile', () => {
...
+      await expect(
+        compileSingleFile(
+          target,
+          new Command(
+            `compile service -s ${process.projectDir}/sasjs/services/common/example.sas`
+          ),
+          'service'
+        )
+      ).toResolve()
+      expect(compileServiceFile.compileServiceFile).toHaveBeenCalled()
```
YuryShkoda, line 143: The fact that `compileJobFile` has been called doesn't prove that a single file has been compiled properly and saved at the correct location.

src/commands/compile/compileSingleFile.ts
allanbowe: if we don't delete the previous content, then it's possible that `sasjs build` will attempt to create build files (.json / .sas) for a previous (project wide) `sasjs compile` - meaning that we couldn't subsequently / consistently `sasjs deploy` a single file

src/commands/compile/spec/compile.spec.ts
saadjutt01: Added `compileJobFile` tests separately.

src/commands/compile/spec/compile.spec.ts
saadjutt01: Added `compileServiceFile` tests separately.

src/commands/compile/compileSingleFile.ts
YuryShkoda: but `sasjsbuild` is also a default location for such files as logs (if another location wasn't provided).

YuryShkoda (approved):

src/commands/compile/compileSingleFile.ts
YuryShkoda: @saadjutt01 you've missed this one

src/commands/compile/internal/compileServiceFile.ts
YuryShkoda: @saadjutt01 you've missed this one

YuryShkoda (changes requested):
not all conversations have been resolved

src/commands/compile/compileSingleFile.ts
allanbowe: correct - all of these items are temporary / transient (not for source control).  Similar to the 'dist' folder (wiped when doing a frontend build).

I do wonder if it would be better if we renamed it as the `sasjstemp` folder.

allanbowe: if the user wants to keep the logs, they can move them afterwards, or redirect them somewhere permanent to begin with

src/commands/compile/compileSingleFile.ts
saadjutt01: it was also fixed with last commit, it has _outdated_ label

src/commands/compile/internal/compileServiceFile.ts
saadjutt01: it was also fixed with last commit, it has _outdated_ label

allanbowe (dismissed):

YuryShkoda (approved):

PR #530, opened 2021-03-03
fix: throw error if target not found with name

### LINKED ISSUE #332: Do not use a fallback target if provided target name was not found
In such a case, a warning `Target with the name 'targetName' was not found in sasjsconfig.json.` should be printed only. CLI shouldn't use a fallback target in such a case.

### LINKED ISSUE #495: when running `sasjs add` and target already exists, provide an option to replace it
Would be nice if the prompt framework would let you choose whether to replace a target (and if so, it would take defaults from it)

image

### LINKED ISSUE #500: Error / Warning on console for compile
there is one thing that keeps coming up every time - there is an ERROR in the console when doing a compile
it should probably be a NOTE

image

The "WARN" could also be removed.
- If a target is not found, the processing should stop.
- If the target is not provided (as in this case) then it should default to the first target with a NOTE

>It's dangerous if I intend to deploy to say, -t dev but I type -t ddev and it defaults to -t prod

saadjutt01:
## Issue

Closes #332
Closes #500 
Closes #495

## Intent

- If target is specified in any `sasjs <command>`, and not found it should stop and do not fall back to first.
- if target name is not specifed in any `sasjs <command`, then fallback message should be _INFO_ not _WARN_
- While `sasjs add target`, if specified `targetName` is found, instead of giving error, show option to update this `target`

## Implementation

Error throw in `findTargetInConfiguration`, if targetName is provided and not found and changed logger warning -> info
`getCommonFields` in `addTarget.ts` is changed and also returns `existingTarget`, added new user input `getAndValidateUpdateExisting`
`add-target.spec.ts` is refactored and added more tests to it.

## Checks

mocked: 
image

server:
image

- [x] Reviewer is assigned.

### DIFF
--- a/src/commands/add/internal/input.ts
+++ b/src/commands/add/internal/input.ts
@@ -10,6 +10,7 @@ import path from 'path'
 import dotenv from 'dotenv'
 import SASjs from '@sasjs/adapter/node'
 import { TargetScope } from '../../../types/targetScope'
+import { CommonFields } from '../../../types/commonFields'
 import {
   findTargetInConfiguration,
   getGlobalRcFile,
@@ -17,14 +18,7 @@ import {
 } from '../../../utils/config'
 import { TargetJson } from '../../../types/configuration'
 
-export async function getCommonFields(): Promise<{
-  scope: TargetScope
-  serverType: ServerType
-  name: string
-  appLoc: string
-  serverUrl: string
-  existingTarget: TargetJson
-}> {
+export async function getCommonFields(): Promise<CommonFields> {
   const scope = await getAndValidateScope()
   const serverType = await getAndValidateServerType()
   const name = await getAndValidateTargetName(serverType)
@@ -63,11 +57,10 @@ async function getAndValidateScope(): Promise<TargetScope> {
     throw new Error(errorMessage)
   })
 
-  if (scope === null || scope === undefined || Number.isNaN(scope)) {
+  if (!scope) {
     throw new Error(errorMessage)
   }
 
-  console.log('Scope: ', scope)
   return scope === 1 ? TargetScope.Local : TargetScope.Global
 }
 
@@ -150,11 +143,10 @@ async function getAndValidateUpdateExisting(
       throw new Error(errorMessage)
     })
 
-    if (choice === null || choice === undefined || Number.isNaN(choice)) {
+    if (!choice) {
       throw new Error(errorMessage)
     }
 
-    console.log('Choice: ', choice)
     return { targetJson, retry: choice === 2 }
   }
 
--- a/src/commands/add/internal/input.ts
+++ b/src/commands/add/internal/input.ts
@@ -15,6 +15,7 @@ import {
   getGlobalRcFile,
   getLocalConfig
 } from '../../../utils/config'
+import { TargetJson } from '../../../types/configuration'
 
 export async function getCommonFields(): Promise<{
   scope: TargetScope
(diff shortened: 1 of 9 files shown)


src/commands/add/internal/input.ts
```diff
@@ -117,6 +123,44 @@ async function getAndValidateTargetName(
...
+      ]
+    ).catch(() => {
+      throw new Error(errorMessage)
+    })
+
+    if (choice === null || choice === undefined || Number.isNaN(choice)) {
+      throw new Error(errorMessage)
+    }
+
+    console.log('Choice: ', choice)
```
YuryShkoda, line 157: do we need this `console.log` ?

src/commands/add/internal/input.ts
```diff
@@ -117,6 +123,44 @@ async function getAndValidateTargetName(
...
+      'Please choose either option 1 or 2.',
+      [
+        { title: '1. Update this target', value: 1 },
+        { title: '2. Go back and create a new target', value: 2 }
+      ]
+    ).catch(() => {
+      throw new Error(errorMessage)
+    })
+
+    if (choice === null || choice === undefined || Number.isNaN(choice)) {
```
YuryShkoda, line 153: maybe we can just check `if (!choice)...`

src/commands/add/spec/add-target.spec.ts
```diff
@@ -15,6 +15,15 @@ import { TargetJson } from '../../../types/configuration'
 import { TargetScope } from '../../../types/targetScope'
 import { createTestMinimalApp, removeTestApp } from '../../../utils/test'
 
+interface CommonFields {
```
YuryShkoda, line 18: it makes sense to put this interface in `types` folder

src/utils/config.ts
```diff
@@ -94,7 +99,7 @@ export async function findTargetInConfiguration(
       : undefined
 
     if (fallBackTargetJson) {
-      process.logger?.warn(
+      process.logger?.info(
         `Target ${targetName || ''} was not found. Falling back to target ${
```
YuryShkoda, line 103: `targetName` will always be an empty string here

src/utils/config.ts
```diff
@@ -117,7 +122,7 @@ export async function findTargetInConfiguration(
     : undefined
 
   if (fallBackTargetJson) {
-    process.logger?.warn(
+    process.logger?.info(
```
YuryShkoda, line 125: `targetName` will always be an empty string here

src/commands/add/internal/input.ts
saadjutt01: We are using same for getting `Scope`
image

Should we change it to logger.Info?

saadjutt01: removed from both places.

src/commands/add/internal/input.ts
saadjutt01: fixed

src/utils/config.ts
saadjutt01: got it, fixed.

YuryShkoda (changes requested):

src/commands/add/spec/add-target.spec.ts
saadjutt01: fixed

src/utils/config.ts
saadjutt01: got it, fixed

YuryShkoda (approved):

PR #581, opened 2021-03-24
fix(doc): support case-insensitve filename for 'doxyfile' + escaping double qoutes & escape characters

### LINKED ISSUE #578: sasjs doc fails when @brief attribute contains double quotes
To reproduce, open this link:  https://gitpod.io/#github.com/sasjs/template_jobs

Then update the job header as per this screenshot and run `sasjs doc`

image

embedded quotes need to be escaped in the DOT file before they can be rendered.

It would also be nice if the DOT file would be saved in the sasjs doc output folder before it is rendered.  This would help with debugging, as well as diagram re-use.

- [ ] sasjs doc supports quotes in the BRIEF when data lineage is enabled
- [ ] DOT output is saved in doc output folder
- [ ] at least one new test case is created

saadjutt01:
## Issue

Closes #578
Support case-insensitive filename for doxygen configuration, currently its `Doxyfile`

## Intent

`sasjs doc`  should not fail when @brief attribute contains double quotes
doxygen configuration filename should be case-insensitive

## Implementation

Updated `createDotFiles.ts` 
- Added try catch for future unknown errors
- Added file creation logger messages

Updated `getDotFileContent.ts `
- embedded quotes escaped before writing to DOT file

## Checks

- [x] Reviewer is assigned.

### DIFF
--- a/src/commands/docs/internal/getDotFileContent.ts
+++ b/src/commands/docs/internal/getDotFileContent.ts
@@ -42,7 +42,9 @@ export async function getDotFileContent(
 
       fileNodes.set(fileName, {
         edges: fileOutputs,
-        label: `${fileName} | ${fileBrief}`.replace(/"/g, '\\"')
+        label: `${fileName} | ${fileBrief}`
+          .replace(/\\/g, '\\\\')
+          .replace(/"/g, '\\"')
       })
     })
   })
--- a/src/commands/docs/internal/getDotFileContent.ts
+++ b/src/commands/docs/internal/getDotFileContent.ts
@@ -42,7 +42,7 @@ export async function getDotFileContent(
 
       fileNodes.set(fileName, {
         edges: fileOutputs,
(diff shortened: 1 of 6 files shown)


krishna-acondy (approved):

src/commands/docs/internal/getDotFileContent.ts
```diff
@@ -42,7 +42,7 @@ export async function getDotFileContent(
 
       fileNodes.set(fileName, {
         edges: fileOutputs,
-        label: `${fileName} | ${fileBrief}`
+        label: `${fileName} | ${fileBrief}`.replace(/"/g, '\\"')
```
allanbowe, line 45: maybe we could also escape backslash here also?

saadjutt01: added

allanbowe (approved):

allanbowe (approved):

PR #582, opened 2021-03-24
feat(lint): Integrate sasjs/lint

### LINKED ISSUE #580: Integrate sasjs/lint into the CLI
The [@sasjs/lint](https://github.com/sasjs/lint) project now provides the capability to identify, for SAS file, whether there are any ERRORs or WARNINGs present and if so, which line number they are on.

We should integrate this into a global `sasjs lint` commant that will search the current, and all parent directories for a `.sasjslint` file.

If found, the settings will be applied to all .sas files in all directories from the .sasjslint file and below.

If issues are found, the return should be non-zero - this will enable us to use the `sasjs lint` command in a git `pre-commit` hook, to prevent questionable SAS code from entering repositories.

saadjutt01:
## Issue

Closes #580 

## Intent

Add new command `sasjs lint`
Looks for folder having `.sasjslint` current to parent directory, 
if found, starts from that directory else from current directory and pass all `.sas` files for linting excluding `node_modules`

## Implementation

Added file `lint/index.ts`, having recursive function `executeLint` from parent folder
Updated `utils/command.ts`, sorted commands and flag values.
~~Added `getDirectoryContainingLintConfig` to `utils/config.ts`~~
Using `lint/utils/getProjectRoot.ts`
Returning none zero error code if any diagnostic of severity level `Error` is found, else printing appropriate message

## Checks

- [x] Reviewer is assigned.


src/utils/config.ts
```diff
@@ -486,6 +486,33 @@ export async function getProjectRoot() {
   return root
 }
 
+export async function getDirectoryContainingLintConfig() {
+  let lintConfigFolder = ''
```
krishna-acondy, line 490: I think it might be best if I exposed this function from the lint library, since the logic is all there already.

saadjutt01: now using `sasjs/lint/utils/getProjectRoot`

src/commands/lint/index.ts
```diff
@@ -0,0 +1,108 @@
...
+        ? chalk.red.bold('Error')
+        : 'Unknown'
+
+    table.push([
+      severity,
+      d.message,
+      `[${d.lineNumber}, ${d.startColumnNumber}]`
+    ])
+  })
+  console.log(table.toString(), '\n')
```
krishna-acondy, line 107: Shouldn't we be using process.logger.log instead?

saadjutt01: updated

krishna-acondy (approved):

src/commands/lint/index.ts
```diff
@@ -0,0 +1,123 @@
...
+/**
+ * Prints Lint Diagnostics as table
+ * @param {string} filePath- the path to file having offences
+ * @param {Diagnostic[]} sasjsDiagnostics- list of offences in particular file
+ */
+const displayDiagnostics = (
+  filePath: string,
+  sasjsDiagnostics: Diagnostic[]
+) => {
+  console.log(`File: ${chalk.cyan(filePath)}`)
```
krishna-acondy, line 79: We can use process.logger here as well, and if we use process.logger.info, we can remove the use of chalk.

allanbowe (dismissed):

src/commands/lint/index.ts
saadjutt01: updated

krishna-acondy (approved):

PR #651, opened 2021-05-06
feat(test-command): implemented macro support, source and output flags

### LINKED ISSUE #634: sasjs test - next steps
With #607 we now have a test flow file generated when running `sasjs compile` that shows coverage for jobs & services.  This flow file also determines the order in which tests are executed (which is currently simple sequential).

To move this forward there are still a few critical features to deliver:

**Macros**

Macros are not deployed (as jobs/services) but the tests will be.  The tests will be deployed in the same way as Jobs/Services, except they will go to an `$(appLoc)|$(sasjsbuild)/tests/macros` directory.  We cannot use the sasjsbuild folder to determine coverage, however we can take a distinct list of files in the sasjsconfig `macroFolders` directory to determine the percentage.

**Folders**

Currently the test files must live in the same directory as the Job/Service being tested.  We also need to support that these files go into a separate directory.  This will go into `testConfig` as:

* testMacroFolders: []
* testJobFolders: []
* testServiceFolders: []

**Filter**

Name of this flag is yet to be determined (if a name is needed at all).  This value will be used to filter the list of tests in the test flow json file.

**csvFile**

configurable output location for test results

**source**

configurable source location for input (flow) json

**logFolder**

Location for the logs from the tests being executed

YuryShkoda:
## Issue

closes #634 

## Intent

- Implemented macro support 
- Implement tests filtering
- Implement source and output flags support

## Checks

Mocked:
image
- [ ] Reviewer is assigned.


src/commands/compile/compile.ts
```diff
@@ -41,6 +47,33 @@ export async function compile(target: Target, forceCompile = false) {
...
+
+    const buildMacroTestFolder = path.join(
+      (await getConstants()).buildDestinationTestFolder,
+      'macros'
+    )
+
+    if (await folderExists(buildMacroTestFolder)) {
+      const macroTestFiles = await (
+        await listFilesAndSubFoldersInFolder(buildMacroTestFolder)
+      ).filter((item: string) => isSasFile(item))
```
krishna-acondy, line 64: Just a little note - when you're applying a filter like this which doesn't need any extra parameters, you can also shorten it to:
```
await (
        await listFilesAndSubFoldersInFolder(buildMacroTestFolder)
      ).filter(isSasFile)
```

YuryShkoda: wow, it's very nice!
thanks!

krishna-acondy (dismissed):
Great job! Very well done. 👏🏽

YuryShkoda:
> Great job! Very well done. 👏🏽

thanks!)))

krishna-acondy (approved):

PR #664, opened 2021-05-11
fix: explicit versions are locked in package.json

### LINKED ISSUE #602: strip trailing slash from appLoc and serverUrl in sasjsconfig.json
If a trailing slash is provided for the appLoc, the CLI (on windows) tries to create a folder with no name.

a trailing slash in the serverUrl appears harmless however it doesn't seem right in the console window.

For appLoc we should strip the trailing slash when we extract it.

For serverUrl we could go on step further and ignore everything after the domain / domain:port

### LINKED ISSUE #606: validate target name before using
if an illegal character (such as an underscore) is used in the target name, this causes a misleading error (credentials not found) in `sasjs deploy`.  Interestingly, underscores work fine for `sasjs cb`.

We should validate the target name when using it.

### LINKED ISSUE #652: sasjs compile filename issue for generated SAS program
To reproduce:

```
sasjs create test -t minimal
cd test/sasjs/services/common
mv getdata.sas get.sasdata.sas

sasjs cb
```

if you open sasjsbuild/myviyadeploy.sas on line 1559 you can see that the filename is `getdata.sas` - but it should be `get.sasdata`

image

### LINKED ISSUE #653: extra security checks:
* update package.json so that all versions are locked by default
*  run `npm audit` as part of the CI build to ensure the dependency tree is clear from warnings

saadjutt01:
## Issue

Closes #653 
Closes #652 
Closes #606
Closes #602

## Intent

Explicitly define versions of all packages in `package.json`
Fix filename for `build.json` and `build.sas` files
Validate target name before using 
Remove leading slash from `apploc` & Extract only domain:(port) from `serverUrl`

## Implementation

- `package.json` is updated as per `package-lock.json` file
- `build.ts` is updated to replace through regex `.sas` at the end of filename: string for both `build.json` and `build.sas`
- Updated codebase to prettier@latest styles
- `src/utils/config.ts` is updated to `sanitizeAppLoc`  + extracted only domain:(port) from `serverUrl`
- `src/utils/config.ts` is updated, error messages placed as `const`

## Checks

- [ ] Reviewer is assigned.


package.json
```diff
@@ -12,7 +12,7 @@
     "copy:files": "copyfiles -u 1 ./src/config.json ./build/ && npm run copy:doxy",
     "copy:doxy": "copyfiles -u 2 src/doxy/* build/doxy/",
     "test": "npm run test:mocked && npm run test:server",
-    "test:server": "jest --silent --config=jest.server.config.js",
```
krishna-acondy, line 15: We should probably keep the silent flag so it doesn't flood the CI build log with unnecessary information.

package.json
```diff
@@ -45,51 +45,53 @@
...
-    "@sasjs/core": "^2.25.0",
-    "@sasjs/lint": "^1.6.1",
-    "@sasjs/utils": "^2.12.1",
-    "btoa": "^1.2.1",
-    "chalk": "^4.1.0",
-    "csv-stringify": "^5.6.1",
-    "dotenv": "^8.2.0",
-    "esm": "^3.2.25",
-    "find": "^0.3.0",
+    "@sasjs/adapter": "2.2.18",
```
krishna-acondy, line 48: The package lock file needs to be regenerated as well.

krishna-acondy (changes requested):

package.json
saadjutt01: fixed

package.json
saadjutt01: `package.json` and `package-lock.json` are synced

krishna-acondy (approved):

Issue #685, opened 2021-05-19
2.22 shows errors on install, currently being fixed


allanbowe:
Current behaviour of 2.22:

![image](https://user-images.githubusercontent.com/4420615/118810688-d8736f80-b8b4-11eb-8d65-e7175a21f9cc.png)

Plan:

* deprecate postinstall
* set ignore-scripts true in the .npmrc

allanbowe:
we can't set ignore-scripts to true in the .npmrc as this also prevents `npm run xxx` commands.   The best we can do is to disable scripts when running `npm i`

So there is no need to deprecate postinstall and we can use it for applying git hooks

allanbowe:
fixed in #687

PR #689, opened 2021-05-19
chore: adding noscripts to .npmrc to prevent malicious scripts from t…

allanbowe:
…hird party dependencies

## Issue

Relates to #685 

## Intent

Adding `ignore-scripts=true` to the `.npmrc` file to avoid the possibility of third party dependencies executing malicious scripts on cli build / dev machines

## Implementation

Added .npmrc

### DIFF
--- a/.npmrc
+++ b/.npmrc
@@ -1,0 +1,1 @@
+ignore-scripts=true


allanbowe:
This totally won't work.. it disables the `npm run` functionality

PR #710, opened 2021-05-27
fix: checking refresh token is expired or not

### LINKED ISSUE #361: sasjs deploy not recognising 401
When running sasjs deploy with invalid (eg expired) credentials the response is misleading:

`An error has occurred when building services Root folder has to be present on the server.`

In fact there should be a response more akin to that in `sasjs run`, eg:

```
Error creating execution session
 Url: https://SAS-SERVER.com
 App loc: /dev/somefolder
 Context: SAS Job Execution compute context
An error has occurred when running your SAS code. {"status":401,"body":"{\"errorCode\":401,\"message\":\"Unauthorized request. Check your credentials(client, secret, access token). Full authentication is 
required to access this resource\",\"details\":[\"path: /compute/contexts\"],\"links\":[],\"version\":2,\"httpStatusCode\":401}"} 
```

### LINKED ISSUE #612: User Friendly message when refresh token expires
We had a couple of support calls recently due to 401 authorisation issues when running `sasjs deploy` or `sasjs run`

The reason was due to the REFRESH_TOKEN expiring

Rather than attempt a connection with an expired token, we should decode the JWT and present a friendly message to the user, eg:

```
The Refresh token for Client $(CLIENT_ID) expired on $(REFRESH_EXPIRY_DTTM).
Please re-authenticate by running sasjs add cred -t $(TARGET_NAME)

Tip - askyour SAS Administrator to generate a client/secret pair with a longer expiry period
```

saadjutt01:
## Issue

Closes #612 
Closes #361 

## Intent

Checks for expiring of _Refresh Token_ if available before using it to refresh _Access Token_

## Implementation

Added new function `isRefreshTokenExpired`
Updated tests

## Checks

- [x] Reviewer is assigned.


src/utils/auth.ts
```diff
@@ -32,6 +32,16 @@ export function isAccessTokenExpiring(token: string) {
   return timeToLive <= 60 * 60 // 1 hour
 }
 
+export function isRefreshTokenExpired(token: string) {
```
allanbowe, line 35: Could we get a tsdoc comment to explain the function?

YuryShkoda (dismissed):

allanbowe (approved):

allanbowe (dismissed):

src/utils/auth.ts
```diff
@@ -32,6 +39,23 @@ export function isAccessTokenExpiring(token: string) {
...
 }
 
+/**
+ * Checks if the Refresh Token is expired or expiring in 30 secs. A default Refresh Token
+ * lasts 30 days.  Once the Refresh Token expires, the user must re-authenticate (provide
+ * credentials in a browser to obtain an authorisation code). 30 seconds is enough time
+ * to make a request for a final Access Token.
+ * @param {string} token- token string that will be evaluated
+ */
+export function isRefreshTokenExpired(token: string): boolean {
```
YuryShkoda, line 49: it makes sense to call this function `isRefreshTokenExpiring` to correspond `isAccessTokenExpiring`

src/utils/config.ts
```diff
@@ -592,7 +597,7 @@ export async function getAccessToken(target: Target, checkIfExpiring = true) {
 
     let authConfig
 
-    if (refreshToken) {
+    if (refreshToken && !isRefreshTokenExpired(refreshToken)) {
```
krishna-acondy, line 600: There's no need for the first part of this check (existence of `refreshToken`) as this is already done within the `isRefreshTokenExpired` function.

saadjutt01: Checking presence of `refreshToken` is for `refreshTokens()` function called from inside of the if block.

src/utils/auth.ts
saadjutt01: done

src/utils/config.ts
saadjutt01: I have toggled if else blocks and removed initial part.

PR #711, opened 2021-05-27
fix: replaced 'fs' with 'fs-extra' completely

saadjutt01:
## Intent

Remove `fs` package 
> This package name is not currently in use, but was formerly occupied by another package. To avoid malicious use, npm is hanging on to the package name, but loosely, and we'll probably give it to you if you want it.

## Implementation

Replaced `fs` with already using `fs-extra`

## Checks

- [ ] Reviewer is assigned.


allanbowe (dismissed):

YuryShkoda (changes requested):
Please move this logic to @sasjs/utils

src/cli.ts
```diff
@@ -23,7 +23,7 @@ import {
   ReturnCode,
   test
 } from './main'
-import { fileExists } from './utils/file'
+import { fileExists } from '@sasjs/utils/file'
```
YuryShkoda, line 26: we should import utils from `@sasjs/utils`

YuryShkoda (changes requested):

src/cli.ts
saadjutt01: done

src/utils/file.js
```diff
@@ -38,159 +36,10 @@ export async function createFolderStructure(folder, parentFolderName = '.') {
...
-    fs.writeFile(fileName, content, function (error) {
-      if (error) {
-        process.logger?.error(`Error creating file ${fileName}`)
-        return reject(error)
-      }
-
-      resolve(content)
-    })
-  })
+export async function listIniFilesInFolder(folderName) {
```
YuryShkoda, line 39: this utility should be in `@sasjs/utils`

saadjutt01: done

PR #742, opened 2021-06-08
feat(vars): added macro variables to flow and job commands

### LINKED ISSUE #492: Enable SAS Macro Variables as Arguments in `sasjs job execute` and `sasjs flow`
Some Viya jobs take parameters as input.  We need to support these in both `sasjs job execute` and `sasjs flow` commands.

We can follow the [approach](https://cli.sasjs.io/context/) taken in `sasjs context` with a `--source` json file, structured as follows:

```
{ "macroVars": { "name1":"value1", "name2":"value2"}}
```

So the `"macroVars"` object would contain a set of name value pairs.  These can be passed as arguments in a similar way to the [mv_jobexecute](https://github.com/sasjs/core/blob/6c9256e0979ab3d58d513ac01917698b8693fe47/viya/mv_jobexecute.sas#L132) macro.

This same JSON object can be embedded in the `flow.json` structure, eg as follows:

```
{
  "name": "simpleFlow",
  "flows": {
    "flow01": {
      "jobs": [
        {
          "location": "/Projects/subfolder/someJob1",
           "macroVars": { "name1":"value1", "name2":"value2"}
        },{
          "location": "/Projects/subfolder/someJob2",
          "macroVars": { "name2":"value3", "name4":"value5"}
        }
      ],
      "predecessors": []
    }
  }
}
```

YuryShkoda:
## Issue

closes #492 

## Intent

- Support `macroVars` in `sasjs flow execute` command.
- Support `macroVars` in `sasjs job execute` command.

## Checks

Mocked: 
image
Server:
 image
- [x] Reviewer is assigned.


src/commands/flow/execute.ts
```diff
@@ -413,16 +417,14 @@ export async function execute(
               )
             }
 
-            let csvData: any = await readFile(csvFileRealPath).catch((err) =>
+            let csvContent = (await readFile(csvFileRealPath).catch((err) =>
               displayError(err, 'Error while reading CSV file.')
```
krishna-acondy, line 421: `csvContent` will be undefined when there's an error caught from `readFile`.
It will make the `csvContent.split` on line 424 fail.

krishna-acondy: You might want to return `''` from inside the catch block to make it safe.

YuryShkoda: good suggestion!

krishna-acondy (commented):
Looks good to me other than the small comment I left.

src/commands/flow/execute.ts
YuryShkoda: done

krishna-acondy (approved):

PR #801, opened 2021-06-29
fix(dependencies): fixed integration with @sasjs/adapter

YuryShkoda:
## Intent

Use the latest version of `@sasjs/adapter`.

## Implementation

Used `jesJobExecutor` in `request` method of the `@sasjs\adapter`.

## Checks

image
Mocked:
image
Server:
image
- [ ] Reviewer is assigned.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1751,8 +1751,9 @@
       }
     },
     "@sasjs/adapter": {
-      "version": "file:../adapter/build/sasjs-adapter-5.0.0.tgz",
-      "integrity": "sha512-w4XdfrHEiXwItE7UDcVJC86rkugmnBjik34FZ5cS1Y9ca8bUOWdg5bTcppjRcSqHB3+r95zMKo10q+TpTYMGLQ==",
+      "version": "2.8.6",
+      "resolved": "https://registry.npmjs.org/@sasjs/adapter/-/adapter-2.8.6.tgz",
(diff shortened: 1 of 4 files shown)


allanbowe (approved):

krishna-acondy (commented):
Are we sure that it should be false and not true? @YuryShkoda @allanbowe

VladislavParhomchik (approved):

allanbowe:
> Are we sure that it should be false and not true? @YuryShkoda @allanbowe



> Are we sure that it should be false and not true? @YuryShkoda @allanbowe

Yes - the changes @medjedovicm made to accommodate earlier `sasjs test` signature changes by @YuryShkoda were only applied to the JES approach (false).  So now `sasjs test` will only run under JES.  Which is fine, as we don't need to inject code.

The two big differences between JES (false) and Compute (true) are:

* JES doesn't require code injection to run jobs (it runs existing jobs using the JES APIs)
* Compute is faster due to the hot session (IF the hot session is used)

@krishna-acondy - I did say earlier we only needed to support Compute for the CLI.  I did not realise that @medjedovicm's changes were only for JES.  So we DO need JES (false) for some cases.  Probably @medjedovicm's changes _should_ have been applied to Compute also, however - we don't actually have a use case for that yet so we may as well leave it for now.

krishna-acondy (approved):

PR #820, opened 2021-07-03
feat: introduced auth command (alias to sasjs add cred)

### LINKED ISSUE #768: sasjs auth - and related issues
With the release of SAS 9 compatibility, we need to do some additional housekeeping:

## sasjs auth

This should provide the _same_ functionality as `sasjs add cred` (it's effectively an alias)

## sasjs add

When running `sasjs add` for a sas9 serverType, at the end of the prompts, we should ask for a username / password

image

We should do the same for `sajs auth`

## sas 9 passwords

If the password doesn't start with `{sas00` (case insensitive) we should WARN:
```
cleartext passwords are a security risk.  Please consider SAS encoded passwords.  For this a server config change is required (AllowEncodedPassword).  More info here: https://support.sas.com/kb/36/831.html
```

If it starts with `{sas001` or `{sas002` we should WARN:
```
This password type can be easily decrypted.  Please consider {sas003} or above.
```

## base64 encoded passwords

To further reduce the risk of accidentally revealing a password, eg when screen sharing, the password should be base64 encoded, with a prefix to ensure that the sensitive part is off screen.  All usage of this password in the sas9 sasjs commands should first attempt to decode this password.  If the decode works, and the prefix is found, we use the decoded version, else the raw version.

sabhas:
## Issue

closes #768

## Intent

- add new command auth
- prompt for username and password when server type is sas9
- save username and password in .env.{target name} if scope is local else store in .sasjsrc file
- encode password
- always decode password before use
 
## Checks

- [x] Reviewer is assigned.


allanbowe:
tests passing on windows

image

allanbowe (approved):

PR #829, opened 2021-07-06
fix(web): apploc with spaces + default value for streamServiceName at root

### LINKED ISSUE #828: sasjs streamed apps with spaces in urls
When the `appLoc` contains spaces, we should pad them with `%20` to make the URL clickable

Also, a missing [`streamServiceName`](https://cli.sasjs.io/sasjsconfig.html#streamConfig_streamServiceName) attribute should not result in an `undefined' in the url (when `streamConfig` is defined in the sasjsconfig root)

image

saadjutt01:
## Issue

Closes #828 

## Intent

If streamConfig is at root level, need to use default value for `streamServiceName: 'clickme'`.
If appLoc has spaces in it, deployed link on console/log must be pad with `%20`, to make the URL clickable

++ changing file extension for SAS9 to create SAS Program bug. 
++ Should convert `./abc.hash.js.map` to `./abc.hash.js-map.sas`

++ Issue for SAS9, unable to load properly base64 encoded for `ICO` in index.html Program.

## Implementation

Added new function to `utils/config`, `getStreamConfig` to always have prioritised config from root & target-specific.
Added `appLoc` to `getLaunchPageCode`, space gets converted to `%20`
Removed `updateFaviconHref`, as it's redundant now.
Updated `sasjsout.ts` to support `ICO` file type.

## Checks

image
Mocked:
image
Server:
image
- [x] Reviewer is assigned.


krishna-acondy (approved):
Looks good to me. There seems to be a merge conflict with the package-lock.json from `main` though.

krishna-acondy (approved):

src/commands/build/internal/getLaunchPageCode.ts
```diff
@@ -1,18 +1,18 @@
...
+const SAS9Code = (streamServiceName: string, appLoc: string) => `
 options notes;
 data _null_;
  format url $256.; 
  rc=METADATA_GETURI("Stored Process Web App",url);
  url=coalescec(url,"localhost/SASStoredProcess");
  putlog "NOTE: SASjs Streaming App Created! Check it out here:" ;
  putlog "NOTE- ";putlog "NOTE- ";putlog "NOTE- ";putlog "NOTE- ";
- putlog "NOTE- " url +(-1) "?_program=&appLoc/services/${streamServiceName}" ;
+ putlog "NOTE- " url +(-1) "?_program=${appLoc}/services/${streamServiceName}" ;
```
allanbowe, line 11: This should remain as `&appLoc` so that the generated build.sas file can be modified by the end user by simply updating the `%let apploc=xxx;` at the start (otherwise it will get hard coded from the sasjsconfig attribute)

src/commands/build/internal/getLaunchPageCode.ts
```diff
@@ -22,7 +22,7 @@ data _null_;
    else url="&systcpiphostname";
  end;
  else url="&systcpiphostname";
- url=cats(url,"/SASJobExecution?_FILE=&appLoc/services/");
+ url=cats(url,"/SASJobExecution?_FILE=${appLoc}/services/");
```
allanbowe, line 25: same issue here (the appLoc needs to be resolved in SAS, so that it can be a parameter at the start of the build.sas program)

allanbowe (changes requested):
Need to address the {appLoc} issue otherwise looks good

VladislavParhomchik (dismissed):

src/commands/build/internal/getLaunchPageCode.ts
saadjutt01: fixed

src/commands/build/internal/getLaunchPageCode.ts
saadjutt01: fixed

VladislavParhomchik (approved):

allanbowe (approved):

PR #851, opened 2021-07-15
Improve handling cases when SAS server did not provide session state

YuryShkoda:
## Intent

Improve handling cases when SAS server did not provide session state.

## Checks

image
Mocked:
image
Server:
image
- [x] Reviewer is assigned.


src/commands/job/execute.ts
```diff
@@ -102,11 +102,28 @@ export async function execute(
       macroVars?.macroVars
     )
     .catch(async (err) => {
+      if (err?.name === 'NoSessionStatus') {
```
krishna-acondy, line 105: I think an (err instanceof NoSessionStatus) check would be nicer here.

src/commands/job/execute.ts
```diff
@@ -102,11 +102,28 @@ export async function execute(
...
+            err.logUrl,
+            100000
+          )
+
+          await saveLog(logData, logFile, jobPath, returnStatusOnly)
+        }
+
+        displayError(err)
+
+        terminateProcess(2)
```
krishna-acondy, line 119: Could we not just throw the error here and let the process.exit get called in main.ts?
I think we should only have process.exit calls at the very top level.

VladislavParhomchik (dismissed):

src/commands/job/execute.ts
YuryShkoda: done

src/commands/job/execute.ts
YuryShkoda: done

sabhas (approved):

PR #892, opened 2021-07-28
tests(*): use single seed app for tests

### LINKED ISSUE #852: Reuse single instance of seed app between tests
Investigate using `git reset` to bring the app back to the original state after each test.

saadjutt01:
## Issue

Closes #852 

## Intent

Use single `minimal-seed-app` / `template-jobs`, across all tests.

## Implementation

Added `pretest.ts` and `posttest.ts`, for creating seed app and deleting after all tests.
Added `resetTestAppAndReuse` to `utils/test.ts` for resetting app state.

## Checks

mocked (Windows PS):
image
 
- [ ] Reviewer is assigned.


posttest.ts
```diff
@@ -0,0 +1,10 @@
+import path from 'path'
+import { APP_NAMES } from './APPS_FOR_TESTING'
+import { deleteFolder, folderExists } from '@sasjs/utils'
+
+Object.entries(APP_NAMES)
+  .map((app) => app[1])
+  .forEach(async (appName) => {
+    const appPath = path.join(__dirname, appName)
+    if (await folderExists(appPath)) await deleteFolder(appPath)
```
krishna-acondy, line 9: `deleteFolder` now uses `rimraf`, so we don't need to check if a folder exists before we delete.

APPS_FOR_TESTING.ts
```diff
@@ -0,0 +1,6 @@
+export const APP_NAMES = {
```
krishna-acondy, line 1: I think I'd rather have this in the `test.ts` module along with all other test-related stuff.

src/utils/test.ts
```diff
@@ -60,6 +61,18 @@ export const createTestMinimalApp = async (
   process.currentDir = process.projectDir
 }
 
+export const resetTestAppAndReuse = async (appName: string) => {
```
krishna-acondy, line 64: I'd probably call this `resetTestApp` since that is what it is doing.

tsconfig.json
```diff
@@ -14,6 +14,8 @@
     "resolveJsonModule": true
   },
   "exclude": [
+    "**/*.spec.ts",
```
krishna-acondy, line 17: This will prevent VS Code from being able to provide IntelliSense for imports and global types in `.spec` and `.server` files.
If we want to exclude them from the final bundle, maybe we should add them to the `.npmignore` if they're not already there?

posttest.ts
saadjutt01: I'll verify on `windows`, if it's required.

src/utils/test.ts
saadjutt01: :), I named it `resetTestApp` but it also sets `process.projectDir` and `process.CurrentDir` as `createTestApp` does.

APPS_FOR_TESTING.ts
saadjutt01: Moved to `test.ts`

tsconfig.json
saadjutt01: I have reverted this, no need to make this change anymore.

posttest.ts
saadjutt01: Updated!

allanbowe:
@saadjutt01  - what is the status on this PR?

saadjutt01:
We need to close this PR, it's not effective as expected to be.

PR #917, opened 2021-08-09
chore(refactor): set constants once to make sure single I/O hits

saadjutt01:
## Issue

Optimise process, like we did for `Target` to fetch once. 
`getConstants` also fetches configuration

## Intent

Restrict `getContants`, to be called once.

## Implementation

Added `Constants` type to `Process`. now fetching constants just once while setting `process.projectDir`.
Updated spec for `generateDocs`, `generateDot` and `initDocs`
Added `await setConstants()` wherever needed, was only missing from specs and `utils/test.ts`
fixes other mocked and server tests as well

## Checks

- [ ] Reviewer is assigned.


src/utils/setConstants.ts
```diff
@@ -0,0 +1,90 @@
+import path from 'path'
+import { getInstalledPath } from 'get-installed-path'
+import { getLocalOrGlobalConfig } from './config'
+import { getAbsolutePath } from './utils'
+
+export const setConstants = async () => {
```
krishna-acondy, line 6: It'd be good if we can have tests for this.

saadjutt01: yea

saadjutt01: Added

PR #1118, opened 2022-02-15
feat(test-command): add @sasjs/server support

### LINKED ISSUE #1108: sasjs test on sasjs/server
We need to be able to run `sasjs test` on [sasjs/server](https://github.com/sasjs/server)!

As a pre-requisite, we would also need sasjs request.

And given how similar it is, we may as well implement sasjs job execute support at the same time.

YuryShkoda:
## Issue

Closes #1108 

## Intent

- Add `@sasjs/server` support to `sasjs test` command

## Implementation

- Added logic to support `SASJS` `ServerType`.
- Added unit test to cover the new functionality.

## Checks

image
Mocked:
image
Server (without `request.spec.server`):
image
- [x] Reviewer is assigned.

### DIFF
--- a/src/commands/testing/spec/testing.spec.server.ts
+++ b/src/commands/testing/spec/testing.spec.server.ts
@@ -21,241 +21,10 @@ import {
 import { contextName } from '../../../utils'
 import dotenv from 'dotenv'
 import { build, deploy } from '../..'
+import SASjs from '@sasjs/adapter/node'
 
 describe('sasjs test', () => {
-  const target: Target = generateTarget()
-  const testUrl = (test: string) =>
-    `${target.serverUrl}/${
-      target.serverType === ServerType.SasViya
-        ? 'SASJobExecution'
-        : 'SASStoredProcess'
-    }/?_program=/Public/app/cli-tests/${
-      target.name
-    }/tests/${test}&_debug=2477&_contextName=${encodeURIComponent(
-      target.contextName
-    )}`
-  const testUrlLink = (test: string) => `"=HYPERLINK(""${testUrl(test)}"")"`
-
-  beforeAll(async () => {
-    await createTestApp(__dirname, target.name)
-    await copyTestFiles(target.name)
-    await build(target)
-    await deploy(target, false)
-
-    process.logger = new Logger(LogLevel.Off)
-  })
-
-  afterAll(async () => {
-    await removeTestServerFolder(`/Public/app/cli-tests/${target.name}`, target)
-    await removeTestApp(__dirname, target.name)
-  })
-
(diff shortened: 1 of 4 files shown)


src/commands/testing/spec/testing.spec.server.ts
```diff
@@ -328,152 +97,392 @@ BRF:1
...
 end_of_record`
 
-    const coverageLcov = await readFile(coverageLcovPath)
+  describe('SASVIYA', () => {
+    const target: Target = generateTarget()
+    const testUrl = (test: string) =>
+      `${target.serverUrl}/${
+        target.serverType === ServerType.SasViya
+          ? 'SASJobExecution'
+          : 'SASStoredProcess'
```
allanbowe, line 106: is this VIYA specific?  If so I guess it doesn't need SASStoredProcess.  If it's generic, it should probably include the Server API route as well?

src/commands/testing/spec/testing.spec.server.ts
```diff
@@ -497,13 +506,313 @@ testteardown,tests/testteardown.sas,sasjs_test_id,not provided,,${testUrlLink(
...
+            '{"test_results":\r\n' +
+            '[\r\n' +
+            '{"TEST_DESCRIPTION":"random test description" ,"TEST_RESULT":"PASS" }\r\n' +
+            ']}',
+          log:
+            '1                                          The SAS System             11:42 Friday, February 11, 2022\r\n' +
+            '\r\n' +
+            'NOTE: Copyright (c) 2016 by SAS Institute Inc., Cary, NC, USA. \r\n' +
+            'NOTE: SAS (r) Proprietary Software 9.4 (TS1M6) \r\n' +
+            '      Licensed to ANALYTIUM LTD- PARTNER LICENCE DATA MANAGER CLIENT, Site 70221619.\r\n' +
```
allanbowe, line 714: This part will break if we ever use a different platform

allanbowe: The date will also be different whenever SAS restarts

allanbowe: Ignore these comments - I did not realise this was a mocked response

allanbowe (approved):

VladislavParhomchik (dismissed):

allanbowe (changes requested):
This is working fine (and SUPER fast!!), the only issue I see is that the console log path says `/SASStoredProcess` when it should say `/SASjsApi/stp/execute`.

This is just a display issue (also a functionality issue as you can't click it).  The test does indeed run.

image

allanbowe (approved):

PR #1144, opened 2022-03-03
test: provided mock responses from server for SASVIYA + SASJS

saadjutt01:
## Issue

no issue to link.

## Intent

Remove dependency of `SASJS` server for specs.
Update specs of testing using `SASJS` server.

## Implementation

Updated, test for both `SASJS` and `SASVIYA`.
Provided mocked responses
Removed deploy steps.
Converted `testing.server.spec.ts` to `testing.spec.ts`

## Checks

- [ ] Reviewer is assigned.

### DIFF
--- a/src/commands/testing/test.ts
+++ b/src/commands/testing/test.ts
@@ -146,9 +146,10 @@ export async function runTest(
   }
 
   await asyncForEach(flow, async (test) => {
+    const pathSepRegExp = new RegExp(path.sep.replace(/\\/g, '\\\\'), 'g')
     const sasJobLocation = [
       target.appLoc,
-      test.replace(sasFileRegExp, '')
+      test.replace(pathSepRegExp, '/').replace(sasFileRegExp, '')
     ].join('/')
 
     const testTarget = test
--- a/src/commands/testing/test.ts
+++ b/src/commands/testing/test.ts
@@ -23,10 +23,10 @@ import {
   decodeFromBase64,
   Target
 } from '@sasjs/utils'
-import SASjs from '@sasjs/adapter/node'
 import path from 'path'
 import chalk from 'chalk'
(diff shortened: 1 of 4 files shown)


src/commands/testing/test.ts
```diff
@@ -258,15 +180,89 @@ export async function runTest(
...
+          displayError(
+            {},
+            `'test_results' not found in server response, to debug click ${testUrl}`
+          )
+
+          printCodeExample()
         }
+      })
+      .catch(async (err) => {
         printTestUrl()
```
saadjutt01, line 263: This major change is nothing, just converted named callback function `handleRes` to unnamed for `.then`.

sabhas: Can you provide a line of explanation about why you changed `handleRes` callback function to unnamed in then block?

src/utils/createSASjsInstance.ts
```diff
@@ -0,0 +1,6 @@
+import SASjs, { SASjsConfig } from '@sasjs/adapter/node'
+
+export const createSASjsInstance = (config: Partial<SASjsConfig>): SASjs => {
+  console.log('actual called')
```
sabhas, line 4: console log statement should be removed.

src/commands/testing/test.ts
saadjutt01: It was redundant, to have named callback function for then.

It was added by me few days ago( in order to have similar behaviour for SASJS) , in order to handle a case, but we configured adapter to return correct response.

src/utils/createSASjsInstance.ts
saadjutt01: Good catch.

sabhas (approved):

PR #1150, opened 2022-03-09
fix: remove trailing slash from serverUrl input

### LINKED ISSUE #1146: trim trailing `/` when pasting severUrl into sasjs cli prompt
When adding `serverUrl` in both `sasjs add` and `sasjs auth` we should strip any trailing slashes when saving to the sasjsconfig.json file.

We should also save to the sasjsconfig file immediately, so that it is saved if the user aborts the prompts.

sabhas:
## Issue

closes #1146 

## Intent

*  remove leading slash from serverUrl input

## Implementation

* `serverUrl.replace(/\/$/, '')`

## Checks

- [ ] Reviewer is assigned.

### DIFF
--- a/src/commands/add/internal/input.ts
+++ b/src/commands/add/internal/input.ts
@@ -89,7 +89,7 @@ export async function getAndValidateServerUrl(target?: TargetJson) {
     target?.serverUrl
   )
 
-  return serverUrl
+  return serverUrl.replace(/\/$/, '')
 }
 
 async function getAndValidateTargetName(


src/commands/add/internal/input.ts
```diff
@@ -89,7 +89,7 @@ export async function getAndValidateServerUrl(target?: TargetJson) {
     target?.serverUrl
   )
 
-  return serverUrl
+  return serverUrl.replace(/\/$/, '')
```
medjedovicm, line 92: In the title, it says the leading slash should be removed, but `$` in regex will match the last character (trailing slash)

allanbowe: aah, that's a bad title.  It should be trailing slash.  Good spot!

allanbowe (approved):

PR #1157, opened 2022-03-11
fix: save log when sasjs request is successful and save empty json fi…

### LINKED ISSUE #1156: save log when sasjs request is successful and save empty json file if no output

sabhas:
…le if no output

## Issue

closes #1156 

## Intent

* Always save log file even when request finishes successfully
* if `res.result` is empty then save empty output file

## Implementation

## Checks

- [ ] Reviewer is assigned.

### DIFF
--- a/src/commands/request/request.ts
+++ b/src/commands/request/request.ts
@@ -107,15 +107,17 @@ export async function runSasJob(
         return
       }
 
-      if (res?.result) res = res.result
-      await writeOutput(outputPathParam, res, isLocal)
+      let output = ''
+      if (res?.result) output = res.result
+      await writeOutput(outputPathParam, output, isLocal)
+      await saveLogFile(sasjs, sasJobLocation, logFile, jobPath)
       result = true
     })
     .catch(async (err: any) => {
       if (err && err.errorCode === 404) {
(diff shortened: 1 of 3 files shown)


src/commands/request/request.ts
```diff
@@ -107,15 +107,17 @@ export async function runSasJob(
...
+      await writeOutput(outputPathParam, output, isLocal)
+      await saveLogFile(sasjs, sasJobLocation, logFile, jobPath)
       result = true
     })
     .catch(async (err: any) => {
       if (err && err.errorCode === 404) {
         displaySasjsRunnerError(configJson.username)
       } else {
-        displayError(err, 'An error occurred while executing the request.')
+        displayError('', 'An error occurred while executing the request.')
```
allanbowe, line 120: what is the reason for not displaying `err` ?

allanbowe: If there is an `err.errorCode` that would be quite relevant.  The problem is that SAS can fail for many reasons, so without those details, it becomes impossible for the end user to debug

medjedovicm (approved):

allanbowe (approved):

PR #1160, opened 2022-03-14
fix: logic causing failiur of request.spec.server

sabhas:
## Issue

none

## Intent

* fix logic that caused the failure of request.spec.server

## Implementation

* if result attribute is present in the response, use that as a response
* if the resultant response is string, try to parse it into json
* if result is not present in response and log is there, remove log from response and use whole response for writing output file.

## Checks

- [x] Reviewer is assigned.

### DIFF
--- a/src/commands/request/request.ts
+++ b/src/commands/request/request.ts
@@ -109,12 +109,13 @@ export async function runSasJob(
 
       let output = ''
 
-      if (res?.result) {
-        res = res.result
-      } else if (res?.log) {
-        delete res.log
-      }
+      if (res?.result) res = res.result
+      // in sasjs request debug is always on and reponse contains log too but it shouldn't be added to output file (*.json),
+      //because for log a separate *.log file is created
+      else if (res?.log) delete res.log
 
+      // sometimes res.result contains log of the request.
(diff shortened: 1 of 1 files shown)


src/commands/request/request.ts
```diff
@@ -108,8 +108,27 @@ export async function runSasJob(
       }
 
       let output = ''
-      if (res?.result) output = res.result
-      await writeOutput(outputPathParam, output, isLocal)
+
+      if (res?.result) {
+        res = res.result
```
YuryShkoda, line 113: shouldn't be wrapped into `{}`

src/commands/request/request.ts
```diff
@@ -108,8 +108,27 @@ export async function runSasJob(
...
+          output = JSON.parse(res)
+        } catch (error) {
+          displayError('Error parsing response. JSON is expected.')
+        }
+      } else {
+        output = res
+      }
+
+      if (!!output) {
+        await writeOutput(outputPathParam, output, isLocal)
```
YuryShkoda, line 129: shouldn't be wrapped into `{}`

allanbowe (approved):

YuryShkoda (approved):

saadjutt01 (approved):

PR #1188, opened 2022-04-11
fix(compile): use @sasjs/core from cli if there is no such module in project folder

### LINKED ISSUE #1113: sasjs compile should not have a hard dependency on the local (project) sasjs/core
There is currently a strange situation in the sasjs/core repo - you can only compile it, if sasjs/core is installed!

The order of checking for macro dependencies should be:

* target `macroFolders`
* root `macroFolders`
* node_modules/@sasjs/core/(subfolders) -> project specific version, **IF** it exists!
* the macros installed along with the sasjs/cli -> version embedded in the CLI

YuryShkoda:
## Issue

Closes #1113 

## Intent

- Use `@sasjs/core` from `@sasjs/cli/node_modules` if there is no such module in project folder.

## Implementation

- Add check if folder exists and use `@sasjs/cli/node_modules` as a fallback location for `sasjsConstants.macroCorePath`.
- Remomed redundant imports.
- Adjusted `setConstants.spec` to cover fallback scenario.

## Checks

- [x] Reviewer is assigned.


sabhas (approved):

saadjutt01 (dismissed):

sabhas (approved):

allanbowe (approved):

src/utils/utils.ts
```diff
@@ -459,3 +459,25 @@ export const isSASjsProject = async () => {
   }
   return false
 }
+
+export const getCliNodeModulePath = async (module: string): Promise<string> => {
+  const cliDepsPath = path.join('@sasjs', 'cli', 'node_modules')
+  const localPath = path.join(
+    process.cwd(),
```
saadjutt01, line 466: It needs to be the root of node application.

src/utils/utils.ts
```diff
@@ -459,3 +459,25 @@ export const isSASjsProject = async () => {
...
+    'node_modules',
+    cliDepsPath,
```
saadjutt01, line 468: It's not necessary to be cli dependency. It may just be module name.

saadjutt01 (changes requested):

PR #1214, opened 2022-06-16
feat(compile): able to sync folder after compile

### LINKED ISSUE #1209: syncFolder feature
We need the ability to add arbitrary content to the sasjsbuild folder (prior to build).  The content could be mocked files (eg JS) or even SAS programs, for developers who wish to add content outside of the regular compile process.

The implementation will be a single attribute at both ROOT and TARGET level as follows:

`"syncFolder": "project/location/myFolder",`

* Any and all files & subfolders inside this will be added inside the sasjsbuild folder
* Any same-named SAS programs (eg `services/common/appinit.sas`) will be overwritten if they exist in the syncFolder
* The `syncFolder` content(s) should be added _after_ the regular compile process (so will replace the compiled version if there is a conflict)
* The ROOT level `syncFolder` is added first, then the TARGET level (if either exist).  So, TARGET files will overwrite ROOT, which will overwrite COMPILED files.
* `sasjs compile` should work for a minimal `sasjsconfig.json` that contains nothing else, eg:

```json
{
  "$schema": "https://raw.githubusercontent.com/sasjs/utils/main/src/types/sasjsconfig-schema.json",
  "syncFolder": "project/location/myFolder",
  "defaultTarget": "server",
  "targets": [
    {
      "name": "sasjs",
      "serverUrl": "",
      "serverType": "SASJS",
      "syncFolder": "target/specificFolder"
    }
}
```

In this case only the contents of `myFolder`  and `specificFolder` added to the `sasjsbuild` folder. 

Tests should include:

- [ ] the syncFolder-only use case above
- [ ] same named SAS program (should be successfully overwritten)
- [ ] file in a new folder, 3 levels deep

The schema json should also be updated.  

Note - this feature should work on ALL server types (SASJS, SAS9 and SASVIYA)

saadjutt01:
## Issue

Fixes #1209 

## Intent

Should be able to copy files from syncFolder of `Configuration.Root` and then of `Target`.

## Implementation

## Checks

- [ ] Reviewer is assigned.

### DIFF
--- a/src/commands/compile/compile.ts
+++ b/src/commands/compile/compile.ts
@@ -37,6 +37,7 @@ import {
   getDestinationJobPath
 } from './internal/getDestinationPath'
 import { getCompileTree } from './internal/loadDependencies'
+import { copySyncFolder } from './internal/copySyncFolder'
 
 export async function compile(target: Target, forceCompile = false) {
   const result = await checkCompileStatus(target, ['tests'])
@@ -101,6 +102,8 @@ export async function compile(target: Target, forceCompile = false) {
   )
 
   await compileWeb(target)
+
+  await syncFolder(target)
 }
 
 export async function copyFilesToBuildFolder(target: Target) {
@@ -376,5 +379,17 @@ async function compileWeb(target: Target) {
           err.toString()
         )
       })
+  }
+}
+
+const syncFolder = async (target: Target) => {
+  const { configuration } = await getLocalOrGlobalConfig()
+
+  if (configuration.syncFolder) {
+    await copySyncFolder(configuration.syncFolder)
+  }
+
+  if (target.syncFolder) {
+    await copySyncFolder(target.syncFolder)
   }
 }
--- a/src/commands/compile/internal/copySyncFolder.ts
+++ b/src/commands/compile/internal/copySyncFolder.ts
@@ -2,13 +2,13 @@ import { copy, folderExists, getAbsolutePath } from '@sasjs/utils'
 
 export const copySyncFolder = async (syncFolder: string) => {
   const { buildSourceFolder, buildDestinationFolder } = process.sasjsConstants
-  const mocksFolder = getAbsolutePath(syncFolder, buildSourceFolder)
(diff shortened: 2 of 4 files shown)


src/commands/compile/compile.ts
```diff
@@ -378,3 +381,15 @@ async function compileWeb(target: Target) {
       })
   }
 }
+
+const syncFolder = async (target: Target) => {
```
YuryShkoda, line 385: In my opinion, this logic should belong to `copySyncFolder` function.

saadjutt01:
Compiling an app, involves process.env.variables, whereas `sasjs/utils` is not introduced to that/ dependent to it.


Our current implement of compile a single file is,
- gathering all the required information and passing it to `loadDependenciesFile` ( sasjs/utils' module )
Which is minimal piece of code from flow of compiling process.

And syncFolder is updating/overwriting entire `sasjsbuild` folder. **It's not part of single file compile.**

We can move `syncFolder` feature to `sasjs/utils`, when we are moving compile process of project.

If we are still motivated to move `copySyncFolder` to `sasjs/utils` in current codebase and again extracting required information(i.e. process env variables) it will be single line function that is `copy(...)`

src/commands/compile/compile.ts
saadjutt01: added separate [comment](https://github.com/sasjs/cli/pull/1214#issuecomment-1158958162) on this PR

src/commands/compile/internal/copySyncFolder.ts
```diff
@@ -0,0 +1,14 @@
+import { copy, folderExists, getAbsolutePath } from '@sasjs/utils'
+
+export const copySyncFolder = async (syncFolder: string) => {
+  const { buildSourceFolder, buildDestinationFolder } = process.sasjsConstants
+  const mocksFolder = getAbsolutePath(syncFolder, buildSourceFolder)
```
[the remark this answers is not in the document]
saadjutt01: updated

allanbowe (approved):

PR #1226, opened 2022-07-07
fix: should not authenticate against sasjs server running in desktop mode

### LINKED ISSUE #1216: `sasjs auth` and `sasjs add` should have a check to avoid client authentication in desktop mode

sabhas:
## Issue

closes #1216

## Intent

* there's no need to authenticate against the target containing `serverType == SASjs` and the server is running in desktop mode.

## Implementation

* Added a utility function that calls the server's info api endpoint and checks if the server is running in server mode or not.

## Checks

- [ ] Reviewer is assigned.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -13,7 +13,6 @@
         "@sasjs/lint": "1.11.2",
         "@sasjs/utils": "2.45.0",
         "adm-zip": "0.5.9",
-        "axios": "0.27.2",
         "chalk": "4.1.2",
         "csv-stringify": "5.6.5",
         "dotenv": "10.0.0",
@@ -2659,6 +2658,7 @@
       "version": "0.27.2",
       "resolved": "https://registry.npmjs.org/axios/-/axios-0.27.2.tgz",
       "integrity": "sha512-t+yRIyySRTp/wua5xEr+z1q60QmLq8ABsS5O9Me1AsE5dfKqgnCFzwiCZZ/cGNd1lq4/7akDWMxdhVlucjmnOQ==",
+      "peer": true,
       "dependencies": {
         "follow-redirects": "^1.14.9",
         "form-data": "^4.0.0"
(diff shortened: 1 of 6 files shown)


src/utils/utils.ts
```diff
@@ -489,3 +490,14 @@ export const getNodeModulePath = async (module: string): Promise<string> => {
   // Return default value
   return ''
 }
+
+export const isServerRunningInServerMode = async (target: Target) => {
+  try {
+    const res = await axios.get(`${target.serverUrl}/SASjsApi/info`)
+    if (res.data.mode === 'server') return true
```
medjedovicm, line 497: Could be written shorter like this:

```
return res.data.mode === 'server'
```

medjedovicm (approved):

allanbowe:
Before we merge this, given the additional dependency, it would be good to know two things:

1.  Will `npm i -g @sasjs/cli` continue to install without errors or warnings
2.  What is the impact on the bundle size, and hence the time to download the CLI?

allanbowe (changes requested):
Before we merge this, given the additional dependency, it would be good to know two things:

1.Will npm i -g @sasjs/cli continue to install without errors or warnings
2. What is the impact on the bundle size, and hence the time to download the CLI?

package.json
```diff
@@ -52,6 +52,7 @@
     "@sasjs/lint": "1.11.2",
     "@sasjs/utils": "2.45.0",
     "adm-zip": "0.5.9",
+    "axios": "0.27.2",
```
YuryShkoda, line 55: You don't need to add `axios` as a dependency because it is already there (brought by `@sasjs/adapter`)

src/utils/utils.ts
```diff
@@ -489,3 +490,14 @@ export const getNodeModulePath = async (module: string): Promise<string> => {
   // Return default value
   return ''
 }
+
+export const isServerRunningInServerMode = async (target: Target) => {
```
YuryShkoda, line 494: I would call it `isSasJsServerInServerMode`

package-lock.json
```diff
@@ -2093,6 +2093,14 @@
         "valid-url": "1.0.9"
       }
     },
+    "node_modules/@sasjs/adapter/node_modules/axios": {
```
YuryShkoda, line 2096: @sabhas , please roll back changes to `package-lock.json`

sabhas:
> 

we are no longer adding any additional dependency in this PR

YuryShkoda (approved):

sabhas:
> Before we merge this, given the additional dependency, it would be good to know two things:
> 
> 1.Will npm i -g @sasjs/cli continue to install without errors or warnings 2. What is the impact on the bundle size, and hence the time to download the CLI?


we are no longer adding any additional dependency in this PR

allanbowe (approved):

PR #1283, opened 2022-11-24
fix: sasjs fs sync with no arguments should deploy per the sasjsconfig

### LINKED ISSUE #1282: `sasjs fs sync` with no arguments should deploy per the sasjsconfig
If `sasjs fs sync` is executed without arguments then we should check the default (or -t specified) target in the project (and then the user-level) config file for the `syncDirectories` array.

If not found, we should say:

> No syncDirectories found

sabhas:
## Issue

closes #1282

## Intent

* If sasjs fs sync is executed without arguments then we should check the default (or -t specified) target in the project (and then the user-level) config file for the syncDirectories array.

## Implementation

* made localFolder and remoteFolder optional and replaced with just local (alias `l`) and remote (alias `r`)
* added a config utility to get all sync directories from target and root of sasjs config
* loop on syncDirectories array returned from `getSyncDirectories`

## Checks

- [ ] Reviewer is assigned.

### DIFF
--- a/src/commands/fs/fsCommand.ts
+++ b/src/commands/fs/fsCommand.ts
@@ -1,6 +1,6 @@
 import { CommandExample, ReturnCode } from '../../types/command'
 import { TargetCommand } from '../../types/command/targetCommand'
-import { displayError, isSASjsProject, saveLog } from '../../utils'
+import { displayError, getSyncDirectories, saveLog } from '../../utils'
 import path, { join } from 'path'
 import {
   createFile,
@@ -11,7 +11,8 @@ import {
   generateProgramToGetRemoteHash,
   generateProgramToSyncHashDiff,
   getRelativePath,
-  generateTimestamp
+  generateTimestamp,
+  SyncDirectoryMap
 } from '@sasjs/utils'
 import { CommandOptions } from '../../types/command/commandBase'
 import { executeCode } from './internal/executeCode'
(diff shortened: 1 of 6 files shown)


src/commands/fs/fsCommand.ts
```diff
@@ -24,11 +25,13 @@ enum FsSubCommand {
 
 const syntax = 'fs <subCommand> [options]'
 const usage = 'Usage: sasjs fs <subCommand> [options]'
-// TODO: update
 const description = 'Handles operations around file system synchronisation'
+const parseOptions = {
+  output: { type: 'string', alias: 'o' }
```
YuryShkoda, line 30: what does `alias: 'o'` mean?

src/commands/fs/fsCommand.ts
```diff
@@ -39,20 +42,32 @@ const compileCommandExamples: CommandExample[] = [
...
+      'sasjs fs sync -l <path/of/folder> -r <path/of/remote/folder> -t <target-name>',
+    description: ''
   }
 ]
 
-const parseOptions = {
-  output: { type: 'string', alias: 'o' }
+const syncCommandParseOptions = {
+  ...parseOptions,
+  local: { type: 'string', alias: 'l' },
```
YuryShkoda, line 69: we should create an enum for aliases

YuryShkoda (dismissed):

allanbowe (approved):

PR #1294, opened 2022-12-13
fix: use correct context name while creating sasjs adapter instance

sabhas:
## Intent

* previously we were not passing the context name while creating sasjs adapter's instance. So, it was using the default context all the time.

## Implementation

* pass target.contextName to the constructor of adapter

## Checks

- [ ] Reviewer is assigned.


allanbowe (approved):
Great spot Sabir, thanks

PR #1328, opened 2023-03-02
fix: use process variable instead of hard coded sasjsbuild

### LINKED ISSUE #1327: `sasjsBuildFolder` not being applied in `sasjs c`
To reproduce:

```
sasjs create demo -t jobs
cd demo
vi sasjs/sasjsconfig.json  
# Add this line:    "sasjsBuildFolder":"test",
sasjs c
ls
```

Expected - the `sasjsbuild` folder should not be listed.

sabhas:
## Issue

closes #1327

## Checks

- [ ] Reviewer is assigned.


src/commands/deploy/internal/executeNonSasScript.ts
```diff
@@ -8,9 +8,10 @@ import { isShellScript, isPowerShellScript } from '../../../utils/file'
 export async function executeNonSasScript(scriptPath: string) {
   const fileExtension = path.extname(scriptPath)
 
+  const { buildDestinationResultsLogsFolder } = process.sasjsConstants
```
allanbowe, line 11: would be good to get a comment here explaining what `buildDestinationResultsLogsFolder` is.  The variable name contains both "build" and "results" so it might be ambiguous as to whether it's `sasjsbuild` or `sasjsresults`

sabhas: good point!
For now, we can add a comment explaining this point but we should rename these constants in the future.

YuryShkoda (approved):

PR #1418, opened 2025-12-11
Add optional fallback of a macroCorePath environment variable

tmoody:
## Issue

1. `@sasjs/cli` has some dependencies on `@sasjs/core`.

When `@sasjs/cli` is used in pipeline testing of an updated `@sasjs/core` package it is preferable to use the updated `@sasjs/core` in place of the cli's `node_modules/@sasjs/core`.

In a clean environment, with no installed npm packages, an invocation of `npx @sasjs/cli` with the `build` argument results in failure as no installation of `@sasjs/core` is returned by the `getNodeModulesPath()` function (npx caches the `@sasjs/core` dependency locally but that temporary `node_modules` path is not being considered).

2. The repository is susceptible to npm supply chain attacks (SHA1-HULUD).

## Intent

1. When `@sasjs/cli` searches for the `@sasjs/core` folder, first read an environment variable that points at the root directory of a chosen local @sasjs/core package. This permits over-riding the @sasjs/core installed as the cli's node_modules dependency.

If no such environment variable is found, or if it is blank, rely instead on node's [`require.resolve()`](https://nodejs.org/api/modules.html#requireresolverequest-options) routine to determine the required package's node_modules path.

2. Prevent automatic execution of `pre` and `post` npm scripts.

## Implementation

1. `src/utils/setConstants.ts` has the job of declaring `sasjsConstants.macroCorePath` as the path to use.
`macroCorePath` is now first set from the value of an environment variable (also named `macroCorePath`). In the even that it is blank a secondary call to `getNodeModulePath('@sasjs/core')` _should_ return a non-blank value unless the `@sasjs/core` dependency has accidentally been removed from the package.json.

A blank value for `sasjsConstants.macroCorePath` is still a possibility and is acceptable for some functionality within `@sasjs/cli`.
However, for a `sasjs` command that requests a `build`, this path is required and processing only continues in `build.ts` if the value is populated, otherwise an informational error message is thrown and the process halts at an early build stage.

2. New file `.npmrc` containing the `ignore-scripts=true` directive configures `npm` to skip automatic processing of associated scripts with a `pre` or `post` prefix when running a given script name (i.e. given `npm install`, no script named `preinstall` or `postinstall` will also be executed in the same submission)

## Checks

- [ ] Unit tests coverage has been increased and a new threshold is set.
- [ ] Development comments have been added or updated.
- [ ] Development documentation coverage has been increased and a new threshold is set.
- [ ] Reviewer is assigned.

### Reviewer checks

- [ ] Any new code is documented.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,1 @@
+* text=auto eol=lf
--- a/.npmrc
+++ b/.npmrc
@@ -1,0 +1,1 @@
+ignore-scripts=true
--- a/package-lock.json
+++ b/package-lock.json
@@ -6,10 +6,9 @@
   "packages": {
     "": {
       "name": "@sasjs/cli",
-      "hasInstallScript": true,
       "license": "ISC",
       "dependencies": {
-        "@sasjs/adapter": "^4.16.0",
+        "@sasjs/adapter": "^4.16.1",
         "@sasjs/core": "4.59.9",
         "@sasjs/lint": "2.4.3",
         "@sasjs/utils": "3.5.2",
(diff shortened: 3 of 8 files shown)


mulahasanovic (dismissed):

mulahasanovic (dismissed):

allanbowe (approved):

allanbowe:
@all-contributors pls add @tmoody for code

PR #1454, opened 2026-07-10
fix: resolve @sasjs/core macros from the project's own node_modules f…

### LINKED ISSUE #1449: `sasjs compile` does not search node_modules/@sasjs/core folders
The order of directory checking is listed here:  https://cli.sasjs.io/artefacts/#sas-macros

currently after adding a new macro to core, running `npm i @sasjs/core` in the project repo does NOT make it available to `sasjs c`  (even though it appears in `node_modules/@sasjs/core`)

Only after adding the core macro the CLI build (ie, releasing a new version of the CLI and upgrading locally) can I access the new core macro

(the workaround would be to add it to the project macro directory, which does work, but ultimately results in 2 copies of the macro)

YuryShkoda:
## Issue

Closes #1449 

## Intent
- `sasjs compile` ignored a project's own `node_modules/@sasjs/core`: `getNodeModulePath('@sasjs/core')` used unscoped `require.resolve`, which walks up from `@sasjs/cli`'s own install location and always found the CLI's *bundled* `@sasjs/core` first. Running `npm i @sasjs/core` in a project to pick up a new/updated macro had no effect until the next `@sasjs/cli` release.

## Implementation

- `getNodeModulePath` now takes an optional `fromDir`, passed as `require.resolve`'s `paths` option, to anchor resolution at a given directory instead of the CLI's own location.
- `setConstants` now resolves `@sasjs/core` in three steps, matching the [documented order](https://cli.sasjs.io/artefacts/#sas-macros):
  1. `macroCorePath` env var (explicit override, unchanged)
  2. the project's own `node_modules/@sasjs/core` (**new**, via `getNodeModulePath('@sasjs/core', process.projectDir)`)
  3. `@sasjs/cli`'s own bundled `@sasjs/core` (existing fallback, now used only when the project has nothing installed)

## Test plan

- [x] New `getNodeModulePath` test in `utils.spec.ts`: real (non-mocked) temp dir with its own `node_modules/<fake-module>`, unresolvable without `fromDir`, resolvable with it
- [x] New/updated `setConstants.spec.ts` tests: project-scoped lookup wins and skips the fallback; fallback is used (in the right order) when the project has nothing installed; updated call-count assertions on the two pre-existing tests
- [x] Ran existing `compile.spec.ts` (21 tests) and `build.spec.ts` (13 tests), which compile real SAS files against real `@sasjs/core` macros — no regressions
- [x] Updated `docs/diagrams/sasjs-compile.md` to describe the new resolution order

## Checks

- [x] Unit tests coverage has been increased and a new threshold is set.
- [x] Development comments have been added or updated.
- [x] Development documentation coverage has been increased and a new threshold is set.
- [x] Reviewer is assigned.

### Reviewer checks

- [x] Any new code is documented.

### DIFF
--- a/src/utils/setConstants.ts
+++ b/src/utils/setConstants.ts
@@ -47,11 +47,6 @@ export const setConstants = async (
   //   1. the `macroCorePath` environment variable, if set - an explicit
   //      override, e.g. for developing against a local @sasjs/core checkout.
   //   2. @sasjs/core installed in the user's own project (process.projectDir).
-  //      This must be checked before falling back to @sasjs/cli's own
-  //      dependency below, otherwise a user who runs `npm i @sasjs/core` to
-  //      pick up a new/updated macro never actually sees it: @sasjs/cli also
-  //      depends on @sasjs/core, so unscoped resolution finds the CLI's own
-  //      bundled copy first and never reaches the project's.
   //   3. @sasjs/cli's own @sasjs/core dependency, as a fallback for projects
   //      that haven't installed @sasjs/core themselves.
   let macroCorePath = (process.env.macroCorePath as string) ?? ''
--- a/src/utils/setConstants.ts
+++ b/src/utils/setConstants.ts
@@ -43,17 +43,24 @@ export const setConstants = async (
   const buildDestinationJobsFolder = path.join(buildDestinationFolder, 'jobs')
   const buildDestinationDbFolder = path.join(buildDestinationFolder, 'db')
   const buildDestinationDocsFolder = path.join(buildDestinationFolder, 'docs')
-  // Edge case: @sasjs/cli has a dependency on @sasjs/core.
-  // When @sasjs/cli is used to submit a test of the @sasjs/core
-  // repo, it is desirable to use that @sasjs/core repo as the dependency rather
-  // than the older version in @sasjs/cli node_modules.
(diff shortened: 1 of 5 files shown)


tmoody (dismissed):

src/utils/setConstants.ts
```diff
@@ -43,17 +43,24 @@ export const setConstants = async (
...
-  // any node_modules installations of @sasjs/core.
+  // Resolution order for the @sasjs/core macros used by 'sasjs compile' etc:
+  //   1. the `macroCorePath` environment variable, if set - an explicit
+  //      override, e.g. for developing against a local @sasjs/core checkout.
+  //   2. @sasjs/core installed in the user's own project (process.projectDir).
+  //      This must be checked before falling back to @sasjs/cli's own
+  //      dependency below, otherwise a user who runs `npm i @sasjs/core` to
+  //      pick up a new/updated macro never actually sees it: @sasjs/cli also
+  //      depends on @sasjs/core, so unscoped resolution finds the CLI's own
+  //      bundled copy first and never reaches the project's.
```
allanbowe: this part should be removed, it's related to previous behaviour

YuryShkoda: removed

tmoody (approved):

